PR #1, opened 2017-05-08
Zooming and Panning Support

Manghi:
I've added capabilities to zoom in and out using `+ | =` or `- | _`, respectively. Panning broke afterwards so I corrected that behavior as well. By default there will be a 100px border which will limit panning capabilities. 

I've also added rudimentary support for setting different resolutions. You can cycle through the one's present by `1 | 2 | 3`.

Finally, I updated the installation instructions to include Mac support.

It is not perfect and there are still some bugs, but this is something to get us going 😄 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,55 @@
 
 Multiplayer Conway's Game of Life!
 
+## Installing Depenencies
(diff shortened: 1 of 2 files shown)


AaronM04:
Merged, thanks! 🥇

PR #2, opened 2017-06-18
Initial Menu UI with 

Manghi:
Groundwork for the navigational menus were the primary focus of this request. Towards the end I started creeping into video/graphics. More focused effort towards modularization too.

General
- Logging support is added. Limit each message to a debug level for cleanliness.
- Explicitly using `rust-sdl2` v.0.25.0 to avoid compat conflicts with other versions
- Corrected panning (though this needs to be reworked again lol)
- I tried to write the menu code so that it was easily extendible in the future for draw() and update()  [went through a few revs of this hehe]

**New** Menu.rs
- Each menu view holds a container of items to be listed
- Current menus are Start Screen, Options, and Video/Audio/Gameplay settings

**New** Video.rs
- Fullscreen support
- Detect supported resolutions

**New** Utils.rs
- ::Graphics:: Common fn to draw text on screen for a given context

### DIFF
--- a/src/video.rs
+++ b/src/video.rs
@@ -1,7 +1,13 @@
 extern crate ggez;
+extern crate env_logger;
 
 use ggez::Context;
-use sdl2::video::FullscreenType;
+use sdl2::video::{FullscreenType, DisplayMode};
+use log::LogLevel;
+
+// Need to replace this with 
+// call SDL_GetDisplayMode() in a loop, SDL_GetNumDisplayModes() times)
+// Enter into a list, point to current
 
 #[derive(Debug, Clone)]
 pub enum ScreenResolution {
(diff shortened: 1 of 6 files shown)


src/video.rs
```diff
@@ -0,0 +1,150 @@
...
+            modes: Vec::new(),
+        }
+    }
+
+    pub fn add_mode(&mut self, new_mode: DisplayMode) {
+            self.modes.push(new_mode);
+    }
+
+    pub fn print_supported_modes(&self) {
+        if log_enabled!(LogLevel::Info) {
```
AaronM04: Why check for `log_enabled!`? Why not just print `info!` lines? I'm assuming if the log is below the Info level, you wouldn't see anything from `info!`, right?

Manghi: You're right, that is completely useless. I pulled it in from the example I saw and didn't think twice about it. I'll remove it in the next PR :smile:

AaronM04: @Manghi It's not too late! Just push a fix to the `interface` branch in your repo. 👍

Manghi: Clean up is live! Thanks for waiting

AaronM04:
I noticed something: if you click on the menu, it draws cells in the grid when you finally start the game. It looks like mouse clicks are handled the same as when the game is started. Also, I hear the fan running, like a lot of stuff is happening, even though it's just a menu.

PR #6, opened 2018-01-05
Supporting ggez 0.4

Manghi:
# Changelog

## Major
1) Migrated from ggez 0.2.2 to 0.4
2) Removed video settings to prepare for rewrite in acccordance with 0.4
3) Added utilities that are not present in 0.4
4) Universe now drawn using SpriteBatching
5) Moved GridView/Zoom/Panning into viewport.rs (Viewport)
6) Moved user input handling into input.rs (InputManager)

## Misc.
1) More values defined as floats to avoid typecasting
2) Updated Copyright to include 2018 (happy new year!)
3)  Updated rust-sdl2 version
5) No bug fixes 😋


src/input.rs
```diff
@@ -0,0 +1,87 @@
...
+use std::collections::VecDeque;
+use ggez::event::{Keycode, MouseButton};
+//use ggez::event::{Button, Axis};
+
+const NUM_OF_INPUTS: usize = 10;
+
+pub enum InputDeviceType {
+    PRIMARY,
+//    GAMEPAD,
+//    CHINCHILLA
```
AaronM04: wut

src/utils.rs
```diff
@@ -16,23 +16,78 @@
...
+            a_min = b_min;
+        }
+        result.x = a_min;
+
+        if b_max < a_max {
+            a_max = b_max;
+        }
+        result.w = a_max - a_min;
+
+        /* veritcal intersection */
```
AaronM04: typo

src/input.rs
Manghi: 😄

src/client.rs
```diff
@@ -982,22 +842,42 @@ impl GridView {
...
+        println!("Adding path {:?}", path);
+        cb = cb.add_resource_path(path);
+    } else {
+        println!("Not building from cargo? Okie dokie.");
+    }
 
-    // save conf to .toml file
-    let mut f = File::create("ggez.toml").unwrap(); //XXX
-    c.to_toml_file(&mut f).unwrap();
+    let ctx = &mut cb.build().unwrap();
```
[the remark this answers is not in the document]
Manghi: Yeah the issue lies within the Mesa driver. 

Basically there are three options since fedora is still on `17.2.4-2.fc27`:
1. Downgrade to version 17.1
2. Upgrade to [Mesa `master`](https://github.com/mesa3d/mesa)
3. [Manually patch](https://github.com/ggez/ggez/issues/194#issuecomment-353228758) version 17.2.*
4. *Secret option* See if you can get [Fedora Rawhide](https://apps.fedoraproject.org/packages/mesa)

src/utils.rs
Manghi: Thanks.

src/client.rs
```diff
@@ -206,21 +206,31 @@ enum Orientation {
...
+                let _ = s.intro_uni.toggle(c as usize, row as usize, player_id);
             }
         }
         Orientation::Diagonal => {
-            for x in 0..width-1 {
-                let _ = s.intro_uni.toggle(col+x, row+x, 0);
+            for x in 0..(width - 1) {
+                let c: isize = col+x;
+                let r: isize = row+x;
+                println!("toggle_line diag: c is {}, r is {}", c, r); //XXX
```
Manghi: Could probably make this `debug!` but I don't think it is needed altogether.

src/client.rs
```diff
@@ -233,10 +243,12 @@ fn init_title_screen(s: &mut MainState) -> Result<(), ()> {
     // 3) Center it
     // 4) get offset for row and column to draw at
 
-//  let resolution = s.video_settings.get_active_resolution();
+    // let resolution = s.video_settings.get_active_resolution();
     let resolution = (config::DEFAULT_SCREEN_WIDTH, config::DEFAULT_SCREEN_HEIGHT);
-    let win_width = resolution.0 as i32 / s.viewport.get_cell_size() as i32;
-    let win_height = resolution.1  as i32 / s.viewport.get_cell_size() as i32;
+    let win_width  = resolution.0 as i32 / s.viewport.get_cell_size() as i32;
```
Manghi: I think we ran into an issue with using the saved `cell_size`, so we had to modify `Conwayste.toml` to get around the issue. I think it would be best if we use the default cell size here instead of getting it from the viewport.

src/client.rs
```diff
@@ -245,8 +257,14 @@ fn init_title_screen(s: &mut MainState) -> Result<(), ()> {
     let logo_width = 9*5 + 9*5;
     let logo_height = letter_height as i32;
 
-    let mut offset_col = ((win_width/2) - (logo_width/2)) as usize;
-    let offset_row = ((win_height/2) - (logo_height/2)) as usize;
+    let mut offset_col = (win_width/2 - logo_width/2) as isize;
+    let offset_row = (win_height/2 - logo_height/2) as isize;
+
+    let toggle = |s_: &mut MainState, col: isize, row: isize| {
```
Manghi: Did not know this could be done.... this is sweet!

src/utils.rs
```diff
@@ -37,7 +37,10 @@ impl Graphics {
         else {
             dst = Point2::new(coords.x, coords.y);
         }
-        draw(_ctx, &mut graphics_text, dst, 0.0)?;
+        let previous_color = graphics::get_color(_ctx);      // store previous color
```
Manghi: I think it would be better to explain why we store the previous color instead of commenting each statement with what it is doing. Just my two cents.

src/menu.rs
```diff
@@ -61,10 +61,10 @@ pub enum MenuItemValue {
 
 #[derive(Debug, Clone)]
 pub struct MenuItem {
-    id:         MenuItemIdentifier,
-    text:       String,
-    editable:   bool,
-    value:      MenuItemValue,
+    pub id:   MenuItemIdentifier,
```
Manghi: Is there any benefit or improvement to using `menu.id` over `menu.get_id()`?

Manghi: I am okay with this :)

src/client.rs
Manghi: We discussed putting constants in their own module which other modules can reference. Thanks for making this change.
`constants.rs`? `consts.rs`?
`definitions.rs`? `defs.rs`?

Manghi:
Approved.

src/utils.rs
AaronM04: Fixed!

PR #22, opened 2019-02-20
fix bin conflict; new bins are client (gui), cli-client, and server

AaronM04:
This is to fix the bin conflict, so we can just go to the root repo and do one of these:
```
cargo run --bin client    # ggez gui
cargo run --bin server
cargo run --bin cli-client  # what was the netwayste client
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,16 +72,22 @@ ln -s ../resources target    # Needed only for the Client
 ## Running the server
 
 ```
-cd netwayste
 cargo run --bin server 0.0.0.0:9000
 ```
 
-## Running the client
+## Running the GUI client
 
 ```
-cd conwayste
 cargo run --bin client
 ```
+Note: at the time of this writing, it does not have network support.
+
+## Running the CLI client
+
+```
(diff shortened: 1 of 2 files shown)


Manghi:
For some reason I had to check out your specific PR revision. I'm not sure why it didn't update when I checked out `fix_bin_conflict`. Here's what I did:
```Bash
$ git clone --recursive https://github.com/conwayste/conwayste
$ cd conwayste
$ git checkout -b aaron/fix_bin_conflict origin/aaron/fix_bin_conflict
$ cargo build (see warning below)

$ cargo run --bin client
error: `cargo run` can run at most one executable, but multiple were specified

$  cargo run --bin cli-client
error: no bin target named `cli-client`

$ cargo run
error: `cargo run` requires that a package only have one executable; use the `--bin` option to specify which one to run
available binaries: client, conway, server, client
```

I needed to cd into `netwayste` and then checkout the specific revision.

Anyways, looks good :+1:  Well done on getting this ecosystem pulled together.

Approved.

PR #33, opened 2019-04-27
Config revamp!

### LINKED ISSUE #25: Configuration redesign
Currently our `ConfigFile` struct has an interface that's hard to use if you just want to test it, without writing files. I noticed it because our unit tests were screwing up my `conwayste.toml`.

AaronM04:
* Save configuration if modified, but don't save more often than 5 seconds (setting in `constants.rs`).
* In `conwayste.toml`, only list settings that are not default settings. The default settings are commented out at the end of the file along with the version of Conwayste that wrote out the file, for user reference.

Fixes #25 .


conwayste/src/client.rs
```diff
@@ -430,9 +436,9 @@ impl MainState {
 
         graphics::set_fullscreen(ctx, config.is_fullscreen() == true);
         vs.is_fullscreen = config.is_fullscreen() == true;
-*/
+        */
 
-        let viewport = viewport::Viewport::new(config.get_zoom_level(), universe_width_in_cells, universe_height_in_cells);
+        let viewport = viewport::Viewport::new(config.get().gameplay.zoom, universe_width_in_cells, universe_height_in_cells);
```
Manghi: I don't know why it just clicked now but Rust's borrowing/mutability rules finally comfort me with this kind of usage. Didn't like the "openness" of it all but now I get it. :+1:

conwayste/src/client.rs
```diff
@@ -819,6 +824,11 @@ impl EventHandler for MainState {
             _ => {}
         }
 
+        if !do_not_quit {
```
Manghi: The 'not' in the name, combined with its Boolean-ness and the fact that it's being negated, is just all so confusing. I was probably the originator :man_facepalming:. Could you change it to something like `let mut quit = false`?

conwayste/src/config.rs
```diff
@@ -18,373 +18,607 @@
...
+        toml::to_string(&default_settings).unwrap()
+    };
+
+    /// A TomlMap for the `DEFAULT_STRING`.
+    static ref DEFAULT_MAP: TomlMap = toml::from_str(DEFAULT_STRING.as_str()).unwrap();
+
+    /// Same as `DEFAULT_STRING` but as a TOML comment and with version string at top.
+    static ref COMMENTED_DEFAULT_STRING: String = {
+        let mut s = String::new();
+        //XXX version string
```
Manghi: Was this a to-do reminder? The `XXX`, since it looks like you do have that below via `version!()`.

conwayste/src/client.rs
AaronM04: Fixed, thanks!

conwayste/src/config.rs
AaronM04: Oops, yeah, forgot to remove.

AaronM04:
Thanks for the review! :grinning:

AaronM04:
@Manghi You'll probably want to merge from master now :)

PR #35, opened 2019-04-30
Set file length to 0 before writing config to avoid corruption!

AaronM04:


AaronM04:
I went ahead and merged this as well as merging into your branch @Manghi . Wouldn't want broken configs! :grimacing:

PR #49, opened 2019-05-31
Fix client/server timeout & keepalive logic

### LINKED ISSUE #41: Killing client does not report that it disconnected on server
Launch a client and connect to a non-lan server and join a room, then type control+c to abruptly abort. The server does not report that the client timed out, and the client is still listed as present in the room.

AaronM04:
Fixes #41


AaronM04:
Thanks!

Issue #50, opened 2019-07-23
Upgrade to ggez 0.5


Manghi:
0.5 was released on July 20, 2019. 

Better to do this now prior as we're still in the early stages of the UI.

See [their changelog](https://github.com/ggez/ggez/blob/master/CHANGELOG.md#050) for details.

Some highlights:

> 
> - Minimum rustc version is now 1.33, rust 2018 edition.
> - We now use winit instead of sdl2 for window creation and events! This removes the last major C dependency from ggez. It also involves lots of minor changes, the full extent of which is still going to evolve.
> - What used to be the text_cached module is now the text module, replacing all the old text stuff with cached text drawing using the glyph_brush crate. This dramatically changes the text API, as well as being faster and more powerful.
> - Apple products are no longer officially supported. They may work fine anyway, and I'll accept PR's for them, but handlin it all myself is too large an investment of time and energy. Sorry. :-( this commit

AaronM04:
Fixed by #51 !

PR #51, opened 2019-07-27
Update conwayste to ggez v0.5

Manghi:
Most of the changes are pretty straightfoward due to new versions of the nalgebra/winit crates and/or changes to ggez's interface. Probably one of the more bigger changes is not needing to store the old color since ggez now associates colors explicitly to draw-ables. Some of the changes (like Text related things) will be handled in the `ui_widgets` branch since its related and foundational. 

Bumped up conwayste to 0.2.0.

### DIFF
--- a/conwayste/src/client.rs
+++ b/conwayste/src/client.rs
@@ -20,7 +20,6 @@ extern crate conway;
 extern crate env_logger;
 extern crate ggez;
 #[macro_use] extern crate log;
-extern crate sdl2;
 #[macro_use] extern crate serde;
 #[macro_use] extern crate version;
 extern crate rand;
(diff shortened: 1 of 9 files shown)


conwayste/src/client.rs
```diff
@@ -409,19 +409,17 @@ fn init_title_screen(s: &mut MainState) -> Result<(), ()> {
...
         let mut config = config::Config::new();
         config.load_or_create_default().map_err(|e| {
             let msg = format!("Error while loading config: {:?}", e);
-            GameError::from(msg)
+            GameError::FilesystemError(msg)
         })?;
 
         let mut vs = video::VideoSettings::new();
-        vs.gather_display_modes(ctx)?;
+        //vs.gather_display_modes(ctx)?;
```
AaronM04: Meh, why not delete this line instead of commenting?

AaronM04:
Fixes #50 .

PR #53, opened 2019-08-25
Ui Widgets

Manghi:
**Purpose:**

Adds the groundwork for Conwayste's user interface.

**Changes:**
* Added error type `ConwaysteError`
* Rewrote input handling as my first approach was over-engineered
  * See `input.rs`
   * Add support for keyboard input
  * Add support to track mouse clicks and dragging.
* A nested module `ui` which adds the concept of a `Widget` trait and the following gui elements which impl it: 
  * See `ui\` folder
  * Button
  * Chatbox
  * Checkbox
  * Label
  * Text input fields
  * Moved all functions from `utils.rs` into `helpers.rs`
* Collection of widgets is a pane
* A layer can contain any number of widgets or panes, and is associated to the current screen the player is viewing
* UI elements have an associated identifier (`WidgetID`) and actionable task (`UserAction`) to help drive behavior for the interacting element
* `UIManager.rs` is intended to facilitate the creation and management of UI elements

**Pending changes:**
- [x] Add a default monospace font
- [x] Ensure a textfield's displayed text fits within its dimensional boundaries
- [x] Refactor all `ggez::Font` usage to reduce memory leaks and code duplication
- [x] Refactor widget creation signature to use a standard API
- [x] Fix [line dragging issue](https://github.com/conwayste/conwayste/pull/53#issuecomment-530858348)
- [x] Create a set of default colors using `Color::from(...)` and `chromatica`
- [x] Remove widget creation from UI manager instantiation
~~Create example with a modal dialog~~ (See issue #80)
- [x] Address all `PR_GATE`'s
- [x] Documentation
- [x] Add unit tests where reasonable
- [x] No new warnings
~~Add `set_position(x,y)` and `set_dimensions(w,h)` to complement  `set_size(rect)`~~ (See issue #81)


conwayste/src/uimanager.rs
```diff
@@ -0,0 +1,140 @@
...
+pub struct UIManager {
+    layers: HashMap<Screen, Vec<Layer>>,
+}
+
+impl UIManager {
+    pub fn new(ctx: &mut Context, config: &Config) -> Self {
+        let font = graphics::Font::default(); // Provides DejaVuSerif.ttf
+        let mut ui_layers = HashMap::new();
+
+        let chat_pane_rect = Rect::new(30.0, 600.0, 300.0, 150.0);
```
AaronM04: I feel like this isn't the right place for this code. The UIManager shouldn't contain code for any particular UI layout -- instead the user of it would do all that.

conwayste/src/client.rs
```diff
@@ -331,7 +331,11 @@ impl MainState {
         color_settings.cell_colors.insert(CellState::Wall,           Color::new(0.617,  0.55,  0.41, 1.0));
         color_settings.cell_colors.insert(CellState::Fog,            Color::new(0.780, 0.780, 0.780, 1.0));
 
-        let font = graphics::Font::default(); // Provides DejaVuSerif.ttf
+        let mut font = graphics::Font::default(); // Provides DejaVuSerif.ttf
```
AaronM04: You should be able to replace this whole section with:
```rust
let font = graphics::Font::default()
       .map_err(|e| GameError::FilesystemError( format!("Could not load or find font. {:?}", e) ))?;
```

conwayste/src/ui/textfield.rs
```diff
@@ -151,19 +149,21 @@ impl TextField {
...
         self.cursor_index = 0;
+        self.visible_start_index = 0;
+        self.visible_end_index = 0;
         self.blink_timestamp = None;
         self.draw_cursor = false;
     }
 
-    /// Advances the cursor to the right by one
-    pub fn inc_cursor_pos(&mut self) {
+    /// Advances the cursor to the right by one character
```
AaronM04: To me, any of the words advance/increment/decrement are confusing when talking about moving the cursor left or right. Just call it move left or move right.

conwayste/src/ui/textfield.rs
```diff
@@ -243,10 +243,10 @@ impl Widget for TextField {
             let text_with_cursor = self.text.clone();
             let text_pos = Point2::new(self.dimensions.x + CURSOR_OFFSET_PX, self.dimensions.y);
 
-            draw_text(ctx, font, Color::from(css::WHITESMOKE), &text_with_cursor, &text_pos, None)?;
+            draw_text(ctx, font, Color::from(css::WHITESMOKE), &text_with_cursor[self.visible_start_index..self.visible_end_index], &text_pos, None)?;
```
AaronM04: It might be a good idea to have variables or constants for the colors of text when highlighted, and normal, so that the actual colors are not mixed everywhere in the code. Just a thought :grin:

conwayste/src/client.rs
Manghi: Ah thanks for the tip!

conwayste/src/client.rs
```diff
@@ -522,15 +389,14 @@ impl MainState {
...
+            inputs:              input::InputManager::new(),
             net_worker:          None,
             single_step:         false,
             arrow_input:         (0, 0),
             drag_draw:           None,
-            win_resize:          0,
-            return_key_pressed:  false,
-            escape_key_pressed:  false,
             toggle_paused_game:  false,
+            power_on_timestamp:  0.0,
```
AaronM04: Can we name this something else? It's not really "powering on" since it's an app.

conwayste/src/client.rs
```diff
@@ -544,34 +410,97 @@ impl EventHandler for MainState {
...
-        match self.stage {
-            Stage::Intro(mut remaining) => {
+        match current_screen {
+            Screen::Intro => {
+                // Any key should skip the intro
+                if self.inputs.key_info.key.is_some() || (self.power_on_timestamp > INTRO_DURATION) {
+                    self.screen_stack.pop();
+                    self.screen_stack.push(Screen::Menu);
+                } else {
+                    self.power_on_timestamp += duration;
```
AaronM04: I expect this will cause rounding errors to accumulate.

conwayste/src/client.rs
```diff
@@ -852,17 +859,25 @@ impl MainState {
...
+            None => panic!("Error in key_down_event! Screen_stack is empty!"),
+        };
 
-        match cur_stage {
-            Stage::Menu => {
+        match current_screen {
+            Screen::Menu => {
                 if cur_menu_state == menu::MenuState::MainMenu {
-                    self.stage = Stage::Run;
+                    // If at 1, then we haven't started the game yet
```
AaronM04: This feels a little hacky to me.

AaronM04:
I noticed the Enter key is no longer switching between paused and running.

AaronM04:
Minor issue: drawing cells now doesn't draw until the mouse up. Previously it started drawing on mouse down.

conwayste/src/client.rs
Manghi: Sure. Bad habits die hard lol

conwayste/src/client.rs
Manghi: How so? They're both `f64`, and I'm not too worried about precision here.

conwayste/src/client.rs
Manghi: Agreed. Would you be opposed if peeked at the screen layer below to see if it was `Screen::Menu` ?

Manghi:
\+ and - will be fixed in next commit.

I'm using the Enter key for chatting now as nearly all PC games use that approach; pausing and resuming the game moved to the `r` key for the time being.

So when changing the behavior of the mouse click, I actually shopped around different user interfaces (Firefox, Gnome Shell to name a few) and found that there wasn't a consistent mouse-down vs mouse-up convention. I personally like it on the up event as that signals the user is committed to the location of the click.

AaronM04:
OK, now that #52 has been merged in, you can do this with your branch:
* merge in `aaron/merge_ui_widgets_into_gui_fixes` into `ui_widgets`, and resolve the conflicts in 3 files.
* merge the latest `master` into `ui_widgets` (I'd expect no conflicts).

Does that make sense?

Manghi:
Yep, thanks for the clear instructions. All went well!

conwayste/src/menu.rs
```diff
@@ -382,31 +381,26 @@ impl MenuSystem {
...
                 ///////////////////////////////
-                let coords = Point2::new(x, y);
+                coords.y += h + 10.0;
                 let res = video_settings.get_resolution();
                 let cur_res_str = format!("{}x{}", res.w, res.h);
 
-                // TODO: color
-                utils::Graphics::draw_text(_ctx, &self.font, self.inactive_color, &cur_res_str,
-                                           &coords, None)?;
+                ui::draw_text(_ctx, Rc::clone(&self.font), self.inactive_color, cur_resolution, &coords)?;
```
AaronM04: `cur_resolution` wasn't defined!

AaronM04: Looks like it should be `cur_res_str`.

conwayste/src/ui/helpe.rs
```diff
@@ -0,0 +1,151 @@
...
+    }
+}
+
+/// Provides a new `Point2` from the specified point a the specified offset.
+pub fn point_offset(p1: Point2<f32>, x: f32, y: f32) -> Point2<f32> {
+    Point2::new(p1.x + x, p1.y + y)
+}
+
+/// Calculates the center coordinate of the provided rectangle
+pub fn center(r: &Rect) -> Point2<f32> {
```
AaronM04: Nice!

conwayste/src/ui/widget.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    // Long-term TODO
+    // The following implementaions tend to use struct member variables so we need to define their own definitions
+    // Refactor into a Sizable trait
+
+    /// Get the size of the widget. Widget must be sizable.
+    fn size(&self) -> Rect;
+
+    /// Get the size of the widget. Widget must be sizable.
```
AaronM04: Comment says `Get`.

conwayste/src/ui/widget.rs
```diff
@@ -0,0 +1,79 @@
...
+    fn set_size(&mut self, new_dimensions: Rect) {
+        ()
+    }
+
+    /// Translate the widget from one location to another. Widget must be sizable.
+    fn translate(&mut self, point: Vector2<f32>);
```
AaronM04: `point` is not a very descriptive name. Maybe `dest` for destination would be better.

conwayste/src/ui/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+
+#[derive(PartialEq, Clone, Copy, Debug)]
+pub enum UIAction {
+    ScreenTransition(Screen),
+    Toggle(ToggleState),
+    EnterText,
+}
+
+#[derive(PartialEq, Debug, Copy, Clone)]
+pub enum WidgetID {
```
AaronM04: Please define this as `pub struct WidgetID(usize)` so that consumers of this API can define their own widgets. Clients can then set constants with the below names for each of the widgets that are being defined.

conwayste/src/ui/label.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    fn draw(&mut self, ctx: &mut Context) -> GameResult<()> {
+
+        let text = Text::new(self.textfrag.clone());
+
+        // If the text is updated, we need to refresh the dimensions of the virtual rectangle bounding it.
+        // unwrap safe b/c if this fails then the game is fundamentally broken and is not in a usable state
+        let recalculated = <Text as Drawable>::dimensions(&text, ctx).unwrap();
+        if recalculated.w != self.dimensions.w  && recalculated.h != self.dimensions.h {
```
AaronM04: Shouldn't this be `||`?

conwayste/src/ui/label.rs
```diff
@@ -0,0 +1,134 @@
...
+        let font: Font = *font;
+
+        let text_fragment = TextFragment::new(string.clone())
+            .scale(*DEFAULT_UI_FONT_SCALE)
+            .color(color)
+            .font(font);
+
+        let text = Text::new(text_fragment.clone());
+        // unwrap safe b/c if this fails then the game is fundamentally broken and is not in a usable state
+        let mut dimensions = <Text as Drawable>::dimensions(&text, ctx).unwrap();
```
AaronM04: It might be better if the dimensions were supplied by the user and not determined by the text size. If the text is too big, something should be done (like truncation from left or right) to make it fit.

This has to do with the `draw` method too.

AaronM04: It looks like ggez already supports wrappable text! https://docs.rs/ggez/0.5.0/ggez/graphics/struct.Text.html#method.set_bounds

conwayste/src/client.rs
AaronM04: OK, fair enough!

conwayste/src/ui/checkbox.rs
```diff
@@ -0,0 +1,175 @@
...
+const LABEL_OFFSET_Y: f32 = -5.0;
+
+/// A standard checkbox widget that can be enabled or disabled via the ToggleState structure.
+impl Checkbox {
+    /// Creates a Checkbox widget.
+    ///
+    /// # Arguments
+    /// * `ctx` - GGEZ context
+    /// * `widget_id` - Unique widget identifier
+    /// * `action` - Unique action identifer
```
AaronM04: What happens if this is not unique?

conwayste/src/client.rs
```diff
@@ -767,170 +923,202 @@ impl MainState {
...
 
     // update
-    fn update_current_screen(&mut self, ctx: &mut Context) {
+    fn update_main_menu(&mut self, ctx: &mut Context) {
         self.process_menu_inputs();
 
-        let is_direction_key_pressed = {
-            self.menu_sys.get_controls().is_menu_key_pressed()
-        };
+        let mouse_point = Point2::new(self.inputs.mouse_info.position.0 as f32, self.inputs.mouse_info.position.1 as f32);
```
AaronM04: This is a code smell. Why not have `mouse_info` be a `Point2`?

conwayste/src/uimanager.rs
```diff
@@ -0,0 +1,165 @@
...
+        }
+        None
+    }
+
+    pub fn get_screen_layers(&mut self, screen:Screen) -> Option<&mut Vec<Layer>> {
+        self.layers.get_mut(&screen)
+    }
+
+    pub fn textfield_from_id(&mut self, screen: Screen, id: WidgetID) -> Option<&mut TextField> {
+        if let Some(layer) = self.get_top_layer_from_screen(screen) {
```
AaronM04: Any particular reason why you're not iterating through all the layers to find one with that ID?

conwayste/src/client.rs
```diff
@@ -413,41 +430,105 @@ impl EventHandler for MainState {
     fn update(&mut self, ctx: &mut Context) -> GameResult<()> {
         let duration = timer::duration_to_f64(timer::delta(ctx)); // seconds
 
-        self.receive_net_updates();
+        self.receive_net_updates(ctx)?;
```
AaronM04: Why does this take a `ctx`?

conwayste/src/client.rs
```diff
@@ -464,13 +545,21 @@ impl EventHandler for MainState {
 
                 self.viewport.update(self.arrow_input);
             }
-            Stage::InRoom => {
+            Screen::InRoom => {
                 // TODO implement
+                if let Some(_k) = self.inputs.key_info.key {
+                    println!("Leaving InRoom to ServerList");
```
AaronM04: Can you please clean up these debug prints?

conwayste/src/client.rs
```diff
@@ -510,89 +609,139 @@ impl EventHandler for MainState {
...
-                } else {
-                    self.input_manager.add(input::InputAction::MouseMovement(x as i32, y as i32));
-                }
-            }
-            Stage::InRoom => {
-                // TODO implement
+    fn mouse_motion_event(&mut self, _ctx: &mut Context, x: f32, y: f32, _dx: f32, _dy: f32) {
+        self.inputs.mouse_info.position = (x, y);
+
+        if self.inputs.mouse_info.mousebutton != MouseButton::Other(0)
```
AaronM04: It would be good to have comments here explaining what's going on.

conwayste/src/client.rs
```diff
@@ -413,41 +430,105 @@ impl EventHandler for MainState {
...
+                                flip_cell(self, cell)
+                            }
+                        }
+                        Some(MouseAction::Drag) => {
+                            if let Some(cell) = self.viewport.get_cell(mouse_pos) {
+                                // Only make dead cells alive
+                                if let Some(cell_state) = self.drag_draw {
+                                    self.uni.set(cell.col, cell.row, cell_state, CURRENT_PLAYER_ID);
+                                } else {
+                                    flip_cell(self, cell)
```
AaronM04: I'm confused about what's going on here.

conwayste/src/error.rs
```diff
@@ -0,0 +1,45 @@
...
+ *  along with libconway.  If not, see <http://www.gnu.org/licenses/>. */
+
+custom_error! {pub ConwaysteError
+    InvalidUIAction {reason: String} = "ConwaysteError->InvalidUIAction->{reason}",
+    NoAssociatedUIAction{reason: String} = "ConwaysteError->NoAssociatedUIAction->{reason}"
+}
+
+pub type ConwaysteResult<T> = ::std::result::Result<T, ConwaysteError>;
+
+impl PartialEq for ConwaysteError {
```
AaronM04: Is this needed?

conwayste/src/input.rs
```diff
@@ -16,145 +16,97 @@
...
-    //MouseRelease(MouseButton),
-    MouseMovement(i32, i32),
-
-//    Gamepad((Button, Axis)),
+#[derive(PartialEq, Clone, Copy, Debug)]
+pub enum MouseAction {
+    Held,
+    Drag,
+    Click,
+    DoubleClick, // For Aaron TODO :)
```
AaronM04: :+1:

AaronM04:
@Manghi nice work! Let me know if there are any parts I can help with without much conflicts.

conwayste/src/constants.rs
```diff
@@ -52,6 +52,12 @@ pub const PIXELS_SCROLLED_PER_FRAME: f32 = 50.0;   // pixels
 pub const CONFIG_FILE_PATH: &str = "conwayste.toml";
 pub const MIN_CONFIG_FLUSH_TIME: Duration = Duration::from_millis(5000);
 
+// user interface
+lazy_static! {
+    pub static ref DEFAULT_UI_FONT_SCALE: Scale = Scale::uniform(20.0);
+    pub static ref DEFAULT_CHATBOX_FONT_SCALE: Scale = Scale::uniform(15.0);
```
AaronM04: Please comment to explain what these numbers mean.

conwayste/src/ui/helpe.rs
```diff
@@ -0,0 +1,151 @@
...
+
+/// Helper function to draw text onto the screen.
+/// Given the string `str`, it will be drawn at the point coordinates specified by `coords`.
+/// An offset can be specified by an optional `adjustment` point.
+///
+/// # Return value
+///
+/// On success, an `Ok((text_width, text_height))` tuple is returned, indicating the width
+/// and height of the text in pixels.
+pub fn draw_text(ctx: &mut Context, font: Rc<Font>, color: Color, text: String, coords: &Point2<f32>) -> GameResult<(f32, f32)> {
```
AaronM04: I'd suggest some new arguments
* bounds: Rect
* an enum to indicate how to handle the text being too big for the width/height (wrap, cut off, etc.)

conwayste/src/ui/button.rs
```diff
@@ -0,0 +1,187 @@
...
+    /// * `widget_id` - Unique widget identifier
+    /// * `action` - Unique action identifer
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use ggez::graphics::Font;
+    /// use ui::Button;
+    ///
+    /// fn new(ctx: &mut Context) -> GameResult<MainState> {
```
AaronM04: Is this a valid example? It returns `GameResult<MainState>`.

(this goes for the other widget examples too)

conwayste/src/ui/button.rs
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    /// Sets the color of the Button's text to the specified ggez `Color`
+    pub fn label_color(mut self, color: Color) -> Self {
+        self.label.set_color(color);
+        self
+    }
+
+    /// Sets the color of the button to the specified ggez `Color`
+    pub fn button_color(mut self, color: Color) -> Self {
```
AaronM04: Why have a setter? The field is public.

conwayste/src/ui/button.rs
```diff
@@ -0,0 +1,187 @@
...
+        //     println!("Hovering over Button, \"{}\" {:?}", self.label.text, point);
+        // }
+    }
+
+    fn on_click(&mut self, _point: &Point2<f32>) -> Option<(WidgetID, UIAction)>
+    {
+        let hover = self.hover;
+        self.hover = false;
+
+        if hover {
```
AaronM04: Why does it matter if we were hovering?

conwayste/src/ui/label.rs
Manghi: I think by specifying the dimensions in the function signature we introduce unnecessary complexity. Speaking from personal experience, it becomes a repetitive game of sizing-the-rectangle. If anything, I would accept leaving the current implementation as the default behavior and adding support for wrap-ability via a builder pattern.

conwayste/src/ui/checkbox.rs
Manghi: If it's not unique, you'll end up interacting with the wrong widget, likely leading to undesired or buggy behavior in the application.

conwayste/src/client.rs
Manghi: I think you meant `position` not `mouse_info`, but yes it should be a `Point2`. Smelly indeed.

conwayste/src/ui/button.rs
Manghi: Was intended to be used with the builder pattern. No use case as of yet... I know what you're going to say so I'll just go ahead and remove it.

conwayste/src/ui/button.rs
Manghi: The hover state check occurs first, prior to any click checking (see `client.rs`, `update_main_menu()` for an example). Hovering ensures the point for the current frame is bounded for a widget.

If the button is part of a `Pane`, then the `Pane` will forward the `on_click()` to its children. Each child should ensure it's in the proper state before confirming the click; the first child producing `Some()` is what the click will register against.

conwayste/src/ui/label.rs
Manghi: Yep.

conwayste/src/client.rs
Manghi: Don't remember. Removing.

conwayste/src/error.rs
Manghi: Are you referring to the `PartialEq` implementation or the general need for a `ConwaysteError`? For the former I followed the `ConwayError` use case so that the same level of support is met.

conwayste/src/ui/button.rs
Manghi: Not intended to be a full example, but merely code to show you how to create and draw a button. It returns `GameResult<MainState>` because that's what `draw` returns, and I'm keeping inline with the ggez API.

conwayste/src/uimanager.rs
Manghi: A user shouldn't be allowed to interact with a widget if it's at a lower layer, so there's no point in iterating over all layers. We just want to check the top-most layer as that is what it shown the user.

conwayste/src/ui/helpe.rs
Manghi: See comment regarding label API.

conwayste/src/ui/checkbox.rs
AaronM04: Doesn't the `widget_id` prevent that?

conwayste/src/error.rs
AaronM04: This compiles without `PartialEq`. I don't see a good reason to keep it.

conwayste/src/ui/chatbox.rs
```diff
@@ -0,0 +1,184 @@
...
+        None
+    }
+
+    fn draw(&mut self, ctx: &mut Context) -> GameResult<()> {
+        let origin = self.dimensions.point();
+
+        if self.hover {
+            // Add in a teal border while hovered. Color checkbox differently to indicate  hovered state.
+            let border_rect = Rect::new(self.dimensions.x-1.0, self.dimensions.y-1.0, self.dimensions.w + 4.0, self.dimensions.h + 4.0);
+            let hovered_border = graphics::Mesh::new_rectangle(ctx, DrawMode::stroke(2.0), border_rect, Color::from(css::TEAL))?;
```
AaronM04: This color would be better as a constant. Maybe `ON_HOVER_COLOR` or that prefixed with `CHATBOX_`.

conwayste/src/input.rs
```diff
@@ -38,8 +39,8 @@ pub struct MouseInfo {
     pub action: Option<MouseAction>,
     pub scroll_event: Option<ScrollEvent>,
     pub down_timestamp: Option<Instant>,
-    pub down_position: (f32, f32),
-    pub position: (f32, f32),
+    pub down_position: Point2<f32>,
+    pub position: Point2<f32>,
```
AaronM04: Thank you! :)

conwayste/src/ui/mod.rs
```diff
@@ -54,22 +54,21 @@ pub enum UIAction {
...
-    MainMenuTestCheckbox,
+#[derive(PartialEq, Eq, Debug, Copy, Clone)]
+pub struct WidgetID(pub usize);
 
-    MainMenuPane1,
-    MainMenuPane1ButtonYes,
-    MainMenuPane1ButtonYesLabel,
-    MainMenuPane1ButtonNo,
-    MainMenuPane1ButtonNoLabel,
+pub const MAINMENU_LAYER1: WidgetID = WidgetID(0x100);
```
AaronM04: this doesn't need to be hex! :grinning:

AaronM04: Also this might be better in constants.rs, because conceptually these constants are part of the API consumer, not the API itself (which is here in the `ui/` folder).

conwayste/src/ui/checkbox.rs
Manghi: It did back when `WidgetID` was an enum. Now it doesn't since its just a usize value, which a user could accidentally re-use. We could keep track of what ID's are in use and panic if they try to re-use an ID during widget creation. LMK what you think.

conwayste/src/client.rs
Manghi: There's a bug in here related to how drag-n-draw clears the final cell, as you pointed out offline.

But the gist of it is that if we're dragging the mouse around, it will set the cell state to alive. If it's just a mouse click (the `else` case, not dragging), then we toggle the cell's state.

conwayste/src/error.rs
Manghi: Indeed it does. We should also remove it for `ConwayError` too then. Removed in one of the newer commits.

conwayste/src/ui/chatbox.rs
Manghi: It's already one of my tasks for the PR before we merge :smile:

conwayste/src/ui/mod.rs
Manghi: Agreed! Will move. I personally prefer defining via hex but it's not a big deal. We'll start at '1' then.

AaronM04:
Note that `Font` is really just a `usize` and is okay to copy. We don't need `Rc<Font>` anywhere.

conwayste/src/uilayout.rs
```diff
@@ -0,0 +1,126 @@
...
+    TextField,
+    UIAction,
+    helpe,
+};
+
+pub struct UILayout {
+    pub layers: HashMap<Screen, Vec<Layer>>,
+}
+
+impl UILayout {
```
AaronM04: I thought the idea would be that everything in the `ui/` folder would be reusable components, without specifying any particular layout of the components. All this stuff setting up widgets and their positions would be better in `client.rs` or somewhere else outside of the `ui/` folder IMHO.

Manghi: This file isn't in the `src/ui/` folder. It's in `src/`.

conwayste/src/uimanager.rs
```diff
@@ -0,0 +1,73 @@
...
+use crate::ui::{
+    Chatbox,
+    Layer,
+    TextField,
+    WidgetID,
+};
+
+pub struct LayoutManager;
+
+impl LayoutManager {
```
Manghi: FWIW I have a long-term goal to make these functions generically reusable, but I decided to leave it for another time. This PR is already aging.

Manghi:
Your updates to `src\ui\textfield.rs` and the introduction of `FontInfo` both look good.

Manghi:
Wrapping seems to be working pretty well on my system + resolution. There were two issues with the character drawing on the right-most border but I worked past that. The rest fell into place pretty smoothly :). I didn't do any additional refactoring yet, and it could use some unit tests.

Manghi:
Wonderful work on the unit tests! Very clear and easy to understand :). Thank you for taking care of that.

AaronM04:
Thanks! You may have noticed I took off the one character addition for padding. I think a better place for that is outside of the reflow_* functions -- maybe like an `inner_size` function only for this widget that returns a smaller Rect than `size` does (padding on all sides), and that width could be used for the characters on a line.

conwayste/src/ui/common.rs
```diff
@@ -0,0 +1,223 @@
...
+ *  <http://www.gnu.org/licenses/>. */
+
+use ggez::graphics::{self, Font, Rect, Text, TextFragment, DrawParam, Color, Scale};
+use ggez::nalgebra::{Point2, Vector2};
+use ggez::{Context, GameResult};
+
+use crate::constants::DEFAULT_UI_FONT_SCALE;
+
+#[macro_export]
+macro_rules! widget_from_id {
```
AaronM04: Nice!

conwayste/src/client.rs
```diff
@@ -1165,6 +1355,81 @@ impl MainState {
...
+
+        if let Some(tf) = TextField::widget_from_screen_and_id(&mut self.ui_layout, Screen::Run, INGAME_PANE1_CHATBOXTEXTFIELD) {
+            if let Some(m) = tf.text() {
+                msg = format!("{}: {}", username, m);
+            }
+            tf.input_state = None;
+            tf.clear();
+        }
+
+        if !msg.is_empty() {
```
AaronM04: I'd prefer returning early in this case.

```rust
if msg.is_empty() {
    return Ok(());
}
```

conwayste/src/uimanager.rs
```diff
@@ -0,0 +1,94 @@
...
+    Label,
+    Layer,
+    Pane,
+    TextField,
+    WidgetID,
+};
+
+// When adding support for a new widget, use this macro to define a routine to search in the
+// `UILayout` under the `Screen` for a `WidgetID` of that type
+macro_rules! widget_from_screen_and_ui {
```
AaronM04: This name is very similar to `widget_from_screen_and_id`. I'd kind of like a different name.

conwayste/src/ui/chatbox.rs
```diff
@@ -0,0 +1,405 @@
...
+    fn reflow_message(msg: &str, width: f32, font_info: &FontInfo) -> VecDeque<(bool, Text)> {
+        let mut texts = VecDeque::new();
+        let max_chars_per_line = (width / font_info.char_dimensions.x) as usize;
+        let mut s = String::with_capacity(max_chars_per_line);
+
+        let mut chars_added = 0;
+        for word in msg.split_whitespace() {
+            let word_chars = Chatbox::count_chars(word);
+
+            // If the word can fit on the next line, but not the current line
```
AaronM04: Much clearer, thanks! :)

conwayste/src/uimanager.rs
```diff
@@ -0,0 +1,94 @@
...
+                return widget.downcast_mut::<TextField>();
+            }
+        }
+        None
+    }
+
+    /// Retreive a Chatbox from its widget ID
+    pub fn chatbox_from_id(ui: &mut UILayout, id: WidgetID) -> Option<&mut Chatbox> {
+        // Not generated using `gen_screen_id_for_widget` as
+        // the Chatbox exists at a fixed position in the layers of Screen::Run
```
AaronM04: I'm confused by this comment.

conwayste/src/client.rs
Manghi: Going to remove the return type since it's not useful after some of these past refactoring. Let me know if further changes are needed.

conwayste/src/ui/layer.rs
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    fn draw(&mut self, ctx: &mut Context) -> GameResult<()> {
+        if self.with_transparency {
+            // TODO: Get resolution from video-settings
+            let mesh = graphics::Mesh::new_rectangle(
+                ctx,
+                DrawMode::fill(),
+                Rect::new(0.0, 0.0, 1920.0, 1080.0),
+                color_with_alpha(css::HONEYDEW, 0.4)
```
AaronM04: Please put this in constants.rs.

conwayste/src/ui/pane.rs
```diff
@@ -0,0 +1,241 @@
...
+    use ggez::graphics::Scale;
+
+    fn create_dummy_pane(size: f32) -> Pane {
+        Pane::new(WidgetID(0), Rect::new(0.0, 0.0, size, size))
+    }
+
+    #[test]
+    fn test_add_widget_to_pane_basic() {
+        let mut pane = create_dummy_pane(1000.0);
+        let font_info = FontInfo {
```
AaronM04: Please de-dup.

AaronM04:
Great work @Manghi ! This is really coming together! :grin:

conwayste/src/ui/layer.rs
```diff
@@ -0,0 +1,288 @@
...
+    use ggez::graphics::Scale;
+
+    fn create_dummy_layer() -> Layer {
+        Layer::new(WidgetID(0))
+    }
+
+    #[test]
+    fn test_add_widget_to_layer_basic() {
+        let mut layer = create_dummy_layer();
+        let font_info = FontInfo {
```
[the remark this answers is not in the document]
Manghi: Good catch!

Manghi:
Added in the error handling for conway/UI that was recently discussed. Removed all warnings, PR gates, and added some more unit tests. I will hold off on implementing the final two checkboxes until these changes are reviewed.

@AaronM04 Take as much time as you need.

AaronM04:
@Manghi I haven't reviewed the code yet, but I do get these errors now:
```
Thu 2019-10-24 07:25:13.334214 [INFO ] - Guessed window DPI factor: 2
Thu 2019-10-24 07:25:13.334262 [DEBUG] - Calculated physical dimensions: 2400x1600
Thu 2019-10-24 07:25:13.482403 [ERROR] - Could not add widget WidgetID(WidgetID(11)) to pane WidgetID(WidgetID(10))! InvalidDimensions { reason: "Widget of WidgetID(WidgetID(11)) is not fully enclosed by Pane of WidgetID(WidgetID(10))" }
Thu 2019-10-24 07:25:13.482441 [ERROR] - Could not add widget WidgetID(WidgetID(12)) to pane WidgetID(WidgetID(10))! InvalidDimensions { reason: "Widget of WidgetID(WidgetID(12)) is not fully enclosed by Pane of WidgetID(WidgetID(10))" }
Thu 2019-10-24 07:25:13.482479 [ERROR] - Could not add widget WidgetID(WidgetID(5)) to pane WidgetID(WidgetID(4))! InvalidDimensions { reason: "Widget of WidgetID(WidgetID(5)) is not fully enclosed by Pane of WidgetID(WidgetID(4))" }
Thu 2019-10-24 07:25:13.482511 [ERROR] - Could not add widget WidgetID(WidgetID(7)) to pane WidgetID(WidgetID(4))! InvalidDimensions { reason: "Widget of WidgetID(WidgetID(7)) is not fully enclosed by Pane of WidgetID(WidgetID(4))" }
Thu 2019-10-24 07:25:13.482541 [ERROR] - Could not add widget WidgetID(WidgetID(1)) to pane WidgetID(WidgetID(4))! InvalidDimensions { reason: "Widget of WidgetID(WidgetID(1)) is not fully enclosed by Pane of WidgetID(WidgetID(4))" }
Thu 2019-10-24 07:25:13.482589 [ERROR] - Could not add widget WidgetID(WidgetID(3)) to pane WidgetID(WidgetID(4))! InvalidDimensions { reason: "Widget of WidgetID(WidgetID(3)) is not fully enclosed by Pane of WidgetID(WidgetID(4))" }
```
And the ServerList, Start Game, etc. buttons on the left side of the menu don't show up.

conwayste/src/error.rs
```diff
@@ -0,0 +1,74 @@
...
+///         println!("I got a SuperError after the downcast");
+///     },
+///     SuperSideKickError(e) => {
+///         println!("I got a SuperSideKickError after the downcast!");
+///     },
+/// );
+/// ```
+///
+#[macro_export]
+macro_rules! handle_error {
```
AaronM04: Tremendous! Good job!

I am a little confused about the default case (when there are no matches). Is there a match arm for that? The example doesn't shed light on how to handle that.

Manghi:
That's my bad. I forgot to actually run it last night lol. Fixes incoming.

conwayste/src/uilayout.rs
```diff
@@ -140,11 +187,45 @@ impl UILayout {
...
-        pane.add(checkbox);
+        match pane_button.set_size(Rect::new(10.0, 130.0, 180.0, 50.0)) {
+            Ok(()) => { },
+            Err(e) => {
+                error!("Could not set size for button during initialization! WidgetID({:?}) {:?}",
+                MAINMENU_TESTBUTTON,
+                e);
+            }
+        }
+        match pane.add(pane_button) {
```
AaronM04: You can just do `pane.add(pane_button).or_else(|e| { ...` here.

Manghi:
Your improvements to `handle_error!()` and textfield input handling are s!
I've moved the remaining two action items into new tickets.

OK2Merge.

PR #82, opened 2019-10-29
Insert patterns with 2-9 and 0; reset with 1

AaronM04:
Change your insert mode with number keys! Works when paused or running.
- [x] insert patterns
- [x] fix bug causing regular cell toggle to not work
- [x] center inserts under cursor
- [x] visible indicator of what will be inserted (0.5 alpha).
- [x] truncate overlay where not writable
- [x] patterns configurable in conwayste.toml
- [x] rotate with Shift-Arrow
- [x] add unit tests for what I changed/added in libconway

### DIFF
--- a/conwayste/src/client.rs
+++ b/conwayste/src/client.rs
@@ -941,12 +941,14 @@ impl MainState {
         // TODO: truncate if outside of writable region
         // TODO: move to new function
         if let Some((ref grid, width, height)) = self.insert_mode {
+            let unwritable_flash_on =  timer::time_since_start(ctx).subsec_millis() % 250 < 125;  // 50% duty cycle, 250ms period
+
             if self.uni_draw_params.player_id < 0 {
                 return Err(format!("Unexpected player ID {}", self.uni_draw_params.player_id).into());
             }
             let player_cell_state = CellState::Alive(Some(self.uni_draw_params.player_id as usize));
-            let mut color = self.color_settings.get_color(Some(player_cell_state));
-            color.a = 0.5;  // semi-transparent since this is an overlay
+            let mut player_color = self.color_settings.get_color(Some(player_cell_state));
+            player_color.a = 0.5;  // semi-transparent since this is an overlay
             if let Some(cursor_cell) = viewport.game_coords_from_window(self.inputs.mouse_info.position) {
                 let (cursor_col, cursor_row) = (cursor_cell.col, cursor_cell.row);
                 grid.each_set(|grid_col, grid_row| {
@@ -956,7 +958,18 @@ impl MainState {
                         // out of range
                         return;
                     }
-                    if let Some(rect) = viewport.window_coords_from_game(viewport::Cell::new(col as usize, row as usize)) {
+                    let (col, row) = (col as usize, row as usize);
+                    if let Some(rect) = viewport.window_coords_from_game(viewport::Cell::new(col, row)) {
+                        let mut color = player_color;
+                        // only error is due to player_id out of range, so unwrap OK here
+                        if !self.uni.writable(col, row, self.uni_draw_params.player_id as usize).unwrap() {
+                            // not writable, so draw flashing red cells
+                            if unwritable_flash_on {
+                                color = *constants::colors::INSERT_PATTERN_UNWRITABLE;
+                            } else {
+                                return;
+                            }
+                        }
                         let p = graphics::DrawParam::new()
                             .dest(Point2::new(rect.x, rect.y))
                             .scale(Vector2::new(rect.w, rect.h))
--- a/conwayste/src/client.rs
+++ b/conwayste/src/client.rs
@@ -1519,23 +1519,17 @@ impl MainState {
     ///
     /// This will return an error if the selected RLE pattern is invalid.
     fn bit_pattern_from_char(&self, keycode: KeyCode) -> Result<(BitGrid, usize, usize), Box<dyn Error>> {
-        // TODO: make this configurable
+        let gameplay = &self.config.get().gameplay;
         let rle_str = match keycode {
-            KeyCode::Key2 => "bob$2bo$3o!",  // SE glider
-            KeyCode::Key3 => "o$obo$2o!",    // SW glider
-            KeyCode::Key4 => "3o$o$bo!",     // NW glider
-            KeyCode::Key5 => "b2o$obo$2bo!", // NE glider
-
-            KeyCode::Key6 => "4bo$5bo$o4bo$b5o!",         // E LWSS
-            KeyCode::Key7 => "bo$o$o$o$o2bo$3o!",         // S LWSS
-            KeyCode::Key8 => "5o$o4bo$o$bo!",             // W LWSS
-            KeyCode::Key9 => "b3o$o2bo$3bo$3bo$3bo$2bo!", // N LWSS
-
-            // https://www.conwaylife.com/wiki/Period-22_glider_gun
-            KeyCode::Key0 => concat!("18b2o25b$19bo7bo17b$19bobo14b2o7b$20b2o12b2o2bo6b$24b3o7b2ob2o6b$24b2o",
-                                     "b2o7b3o6b$24bo2b2o12b2o2b$25b2o14bobob$35bo7bob$43b2o2$2o23bo19b$bo21b",
-                                     "obo19b$bobo13b3o4b2o19b$2b2o3bo8bo3bo24b$6bob2o6bo4bo23b$5bo4bo6b2obo",
-                                     "9bo14b$6bo3bo8bo3b2o6bo13b$7b3o13bobo3b3o13b$25bo19b$25b2o!"),
+            KeyCode::Key2 => &gameplay.pattern2,
+            KeyCode::Key3 => &gameplay.pattern3,
+            KeyCode::Key4 => &gameplay.pattern4,
+            KeyCode::Key5 => &gameplay.pattern5,
+            KeyCode::Key6 => &gameplay.pattern6,
+            KeyCode::Key7 => &gameplay.pattern7,
+            KeyCode::Key8 => &gameplay.pattern8,
+            KeyCode::Key9 => &gameplay.pattern9,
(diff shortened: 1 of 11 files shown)


AaronM04:
@manghi when you get a chance, try pressing a number other than 1 and then clicking in the writable region.

conwayste/src/client.rs
```diff
@@ -960,6 +979,15 @@ impl MainState {
             return Ok(());
         }
 
+        if keycode == KeyCode::Key1 {
```
Manghi: Should move these into the match underneath

```rust
KeyCode::Key1 => {
    ...
}
k if k >= KeyCode::Key2 && k <= KeyCode::Key0 => {
    ...
}
KeyCode::Return => {
    ...
}
...
```

conwayste/src/client.rs
```diff
@@ -1399,6 +1427,32 @@ impl MainState {
...
+
+            // https://www.conwaylife.com/wiki/Period-22_glider_gun
+            KeyCode::Key0 => concat!("18b2o25b$19bo7bo17b$19bobo14b2o7b$20b2o12b2o2bo6b$24b3o7b2ob2o6b$24b2o",
+                                     "b2o7b3o6b$24bo2b2o12b2o2b$25b2o14bobob$35bo7bob$43b2o2$2o23bo19b$bo21b",
+                                     "obo19b$bobo13b3o4b2o19b$2b2o3bo8bo3bo24b$6bob2o6bo4bo23b$5bo4bo6b2obo",
+                                     "9bo14b$6bo3bo8bo3b2o6bo13b$7b3o13bobo3b3o13b$25bo19b$25b2o!"),
+            _ => "", // unexpected
+        };
+        let pat = Pattern(rle_str.to_owned());
+        let (width, height) = pat.calc_size().unwrap();  // unwrap OK because calc_size should never fail on valid RLE
```
Manghi: If it does fail then we've hardcoded an RLE incorrectly, or libconway produced an incorrect one, and we'd want to fix that prior to a release. 
:+1:

libconway/src/rle.rs
```diff
@@ -164,6 +164,38 @@ fn digits_to_number(digits: &Vec<char>) -> ConwayResult<usize> {
...
+    /// Is `ch` a valid character?
+    fn is_valid(_ch: char) -> bool { true }
+
+    /// Width in cells
+    fn width(&self) -> usize { self.col + 1 }
+
+    /// Height in cells
+    fn height(&self) -> usize { self.row + 1 }
+
+    fn get_run(&self, _col: usize, _row: usize, _visibility: Option<usize>) -> (usize, char) { unimplemented!("PatternSize is write-only"); }
```
Manghi: Would make this into an actual block given that it's a long line.

conwayste/src/client.rs
```diff
@@ -527,29 +530,44 @@ impl EventHandler for MainState {
                     ).unwrap(); // OK to call unwrap here because there is an else match arm (all errors handled)
                 }
 
-                if self.inputs.mouse_info.mousebutton == MouseButton::Left {
+                // TODO: move this into process_running_inputs
+                if self.inputs.mouse_info.mousebutton == MouseButton::Left && self.inputs.mouse_info.action == Some(MouseAction::Click) {
```
Manghi: Bug here. Only `MouseAction::Click`'s are allowed through, which creates dead code in the `else` branch when `self.insert_mode == None`. Result is that I cannot directly draw single cells.

I think you want to move the second-half of this conditional to line 537-539 so that only clicks will drop patterns.

AaronM04: Thank you for finding this!

libconway/src/rle.rs
AaronM04: Fixing this.

conwayste/src/client.rs
AaronM04: Fixing this.

conwayste/src/client.rs
AaronM04: I agree. I'll move this.

conwayste/src/client.rs
AaronM04: Fixed. Now we will see something like this in the logs if the RLE is bad:
```
Wed 2019-10-30 21:30:28.226259 [ERROR] - Received unexpected error from process_running_inputs(). "Invalid pattern bound to keycode Key8: ConwayError->InvalidData->Premature termination at Some(12)"
```

The 12 is a character number.

Manghi: Beautiful!

Manghi:
Everything looks great so far!

AaronM04:
Any parts of the overlay pattern outside the player's writable area will now be flashing red.

conwayste/src/client.rs
```diff
@@ -628,13 +628,41 @@ impl EventHandler for MainState {
...
+                        out_result = Err(e);
+                    },
+                    else => |e| {
+                        error!("Error from draw_intro: {}", e);
+                        do_return = true;
+                    }
+                };
+                if do_return {
+                    return out_result;
+                }
```
Manghi: I'm not seeing the point in having either `out_result` or `do_return`.

1) `out_result` is assigned to in the `GameError`, but the assigned value is not used afterwards.

2) When `do_return` is true, the `out_result` will always be `Ok(())`, so why have the variable at all?

Same goes for lines 651 to 665  in the case for `Screen::Run`.

conwayste/src/client.rs
```diff
@@ -904,20 +934,53 @@ impl MainState {
...
+            let mut color = self.color_settings.get_color(Some(player_cell_state));
+            color.a = 0.5;  // semi-transparent since this is an overlay
+            if let Some(cursor_cell) = viewport.game_coords_from_window(self.inputs.mouse_info.position) {
+                let (cursor_col, cursor_row) = (cursor_cell.col, cursor_cell.row);
+                grid.each_set(|grid_col, grid_row| {
+                    let col = (grid_col + cursor_col) as isize - width as isize/2;
+                    let row = (grid_row + cursor_row) as isize - height as isize/2;
+                    if col < 0 || row < 0 {
+                        // out of range
+                        return;
```
Manghi: Silently return? Why no error?

libconway/src/grids.rs
```diff
@@ -199,6 +199,22 @@ impl BitGrid {
             }
         }
     }
+
+    /// Calls callback on each bit that is set (1). Callback receives (col, row).
+    pub fn each_set<F: FnMut(usize, usize)>(&self, mut callback: F) {
+        for row in 0 .. self.height() {
+            let mut col = 0;
+            for col_idx in 0 .. self.width_in_words() {
+                for shift in (0..64).rev() {
```
Manghi: `64` is a magic number. I know it won't change any time soon but giving it a name helps with *the why*. This is used all over libconway code.

conwayste/src/client.rs
Manghi: This returns out of the closure. Wasn't quite apparent at first.

Manghi:
Awesome job with the overlay.

What do you think about using the same alpha channel value for non-writable regions? It can be a bit disorienting with the overlay blinking a fully opaque red while the second player (color currently red) also has active cells in the  area.

AaronM04:
> Awesome job with the overlay.
> 
> What do you think about using the same alpha channel value for non-writable regions? It can be a bit disorienting with the overlay blinking a fully opaque red while the second player (color currently red) also has active cells in the area.

Good catch! I fixed it.

libconway/src/grids.rs
AaronM04: What do you think about adding a `bits!` macro that will count the bits in the given type (at compile time). Then the `64` could be replaced with `bits!(u64)`.

It can be done at compile time because both `max_value` and `count_ones` are const fns:
```
$type::max_value().count_ones() as usize
```

Manghi: Great suggestion! Especially the compile-time aspect. :vulcan_salute:

Manghi:
So far it all looks good. :+1:

conwayste/src/client.rs
AaronM04: How about this. OK now?

Manghi: This makes much more sense about the intended purpose. Thanks. 

The only thing I see now is that when `do_return` is `true`, we bail out of the draw routine early, specifically skipping the call to `graphics::present()`. If the error persists between event loops, then this will result in the game appearing to have locked-up because we're aborting the frame updates. Doesn't seem like what we would want to happen. I don't think we should return early; printing the error should be enough for debugging; deciding what we do when we get the error is a separate discussion.

AaronM04: Maybe it would be better to not return for `draw_universe` errors. I updated this so that `do_return` is removed.

Manghi:
Good stuff on the keymods. Previously didn't allow more than one modifier per frame. :+1:

AaronM04:
@Manghi this is ready for you to test! :grinning:

libconway/src/grids.rs
```diff
@@ -224,14 +224,50 @@ impl BitGrid {
...
         let mut new = BitGrid::new(new_width_in_words, new_height);
+        for row in 0 .. height {
+            let new_col = match rotation {
+                Rotation::CCW => row,
+                Rotation::CW => height - row - 1,
+            };
+            let new_col_idx = new_col/64; // the column index in the new grid
+            let mut col = 0;
+            'rowloop:
+            for col_idx in 0 .. self.width_in_words() {
```
Manghi: I think the `'rowloop:` belongs on the same line as the for statement if we're following rustfmt conventions. Do correct me if I am wrong about this. I'm basing it off of `'label: loop {}`

Manghi:
Everything looks great :smile: . Very straightforward to read as well. Merge away.

One comment (which we can address in a ticket if you deem it necessary) is: should the transparent-insert-pattern blink red when we're hovering over a wall, even though it is in the writable region?

AaronM04:
@Manghi thanks for looking! I do want to make it blink red if there's a wall there. I want to do it in this PR. Maybe we can knock it out on Thursday?

Issue #79, opened 2019-10-23
Change `Layer` design to Z-index instead of a virtual container


Manghi:
The current implementation via changes found in #53 is design based on the following diagram. Layers physically  `Widget`s, which then makes it rather tedious when you want to interact with a specific widget, such as Pane 2's button. 
```
+----------------------------------------+
| Layer                                  |
|   +----------+      +-----------+      |
|   |Pane 1    |      |Pane 2     |      |
|   |          |      |           |      |
|   | +------+ |      | +-------+ |      |
|   | |Button| |      | |Button | |      |
|   | +------+ |      | +-------+ |      |
|   +----------+      +-----------+      |
|                                        |
|    +--------------------+              |
|    |TextField           |              |
|    +--------------------+              |
|                                        |
+----------------------------------------+
```

New thought is to remove the concept of a Layer and introduce the concept of a Z-depth (or z-index like in css). The Z-index would denote the draw order of the widget, completely removing the need for a Layer construct. Should simplify things a little.

Manghi:
Merged to master. :heavy_check_mark: 
Closing.

PR #83, opened 2019-11-05
Change layering design to use z orders

Manghi:
See #79 for information on the issue.

- [x] Change backend from a double vector to single, ordered list
- [x] Updated API for new backend design
- [x] Update widget interaction API (draw, on_hover, etc.)
- [x] Fill in transparency support
- [x] Update unit tests
- [x] No new warnings
- [x] Replace `id_cache` with a recursive search for id's
  - [x] Change design to use a tree data structure
  - [ ] Migrate `Label` from `Button` and other widgets to child node of the widget
- [x] Reassess `focused_ids` design 
- [ ] Rename `layer.rs` to `layering.rs` after all changes are complete to preserve history


Manghi:
@AaronM04 Ready for review at your leisure!

conwayste/src/ui/common.rs
```diff
@@ -49,6 +45,35 @@ macro_rules! widget_from_id {
...
+///
+/// # Examples
+/// ```rust
+/// let widget = layer.get_widget_mut(WidgetID(0));
+/// let textfield = downcast_widget!(mut widget, TextField);
+/// textfield.enter_focus()
+/// ```
+///
+#[macro_export]
+macro_rules! downcast_widget {
```
AaronM04: Nice work -- the code looks cleaner with this macro! :grin: 

I think it looks a little strange to have `mut x` where you cannot do `x = <new value>` which can be the case with this macro.

It might be better as `downcast_widget!` and `downcast_widget_mut!`.

conwayste/src/ui/layer.rs
```diff
@@ -31,137 +31,251 @@ use super::{
...
+
+    /// Removes a widget belonging to the layering
+    ///
+    /// # Error
+    /// A WidgetNotFound error can be returned if a widget with the `widget_id` does not exist
+    /// in the internal list of widgets.
+    pub fn _remove_widget(&mut self, widget_id: WidgetID) -> UIResult<()> {
+        let removal_index = self.widget_list
+            .iter()
+            .position(|widget| widget_id >= widget.id())
```
AaronM04: What is the reason this is `>=`?

conwayste/src/ui/ui_errors.rs
```diff
@@ -15,11 +15,33 @@
...
     InvalidDimensions {reason: String} = "UIError::InvalidDimensions({reason})",
     WidgetNotFound {reason: String} = "UIError::WidgetNotFound({reason})",
     InvalidAction {reason: String} = "UIError::InvalidAction({reason})",
-    ActionRestricted{reason: String} = "UIError::ActionRestricted({reason})"
+    ActionRestricted{reason: String} = "UIError::ActionRestricted({reason})",
+    WidgetIDCollision{reason: String} = "UIError::WidgetIDCollision({reason})",
+    GgezError{reason: String} = "UIError::GgezError({reason})"
 }
 
 pub type UIResult<T> = Result<T, Box<UIError>>;
```
AaronM04: Just a note -- we will probably need to replace this with `CwResult` which would be `Result<T, Box<dyn Error>>`. There is no benefit to having a concrete type in `Box<...>` here.

Not something for this PR though -- it'll be more far reaching changes.

conwayste/src/ui/layer.rs
```diff
@@ -31,137 +31,251 @@ use super::{
...
-    fn on_click(&mut self, point: &Point2<f32>) -> Option<(WidgetID, UIAction)> {
-        for w in self.widgets.iter_mut() {
-            let ui_action = w.on_click(point);
+    pub fn on_click(&mut self, point: &Point2<f32>) -> Option<(WidgetID, UIAction)> {
+        let highest_z_index = self.peek_z_index();
+
+        for widget in self.widget_list
+            .iter_mut()
+            .filter(|widget| widget.z_index() == highest_z_index)
+        {
```
AaronM04: Please put this at the end of the line above.

conwayste/src/ui/layer.rs
```diff
@@ -31,137 +31,251 @@ use super::{
...
-    fn on_drag(&mut self, original_pos: &Point2<f32>, current_pos: &Point2<f32>) {
-        for w in self.widgets.iter_mut() {
-            w.on_drag(original_pos, current_pos);
+    pub fn on_drag(&mut self, original_pos: &Point2<f32>, current_pos: &Point2<f32>) {
+        let highest_z_index = self.peek_z_index();
+
+        for widget in self.widget_list
+            .iter_mut()
+            .filter(|widget| widget.z_index() == highest_z_index)
+        {
```
AaronM04: Same here.

conwayste/src/ui/layer.rs
```diff
@@ -31,137 +31,251 @@ use super::{
...
+    pub fn on_hover(&mut self, point: &Point2<f32>) {
+        let highest_z_index = self.peek_z_index();
 
-    fn on_hover(&mut self, point: &Point2<f32>) {
-        for w in self.widgets.iter_mut() {
-            w.on_hover(point);
+        for widget in self.widget_list
+            .iter_mut()
+            .filter(|widget| widget.z_index() == highest_z_index)
+        {
```
AaronM04: And here :)

conwayste/src/ui/pane.rs
```diff
@@ -88,6 +90,19 @@ impl Pane {
         Ok(())
     }
 
+    pub fn get_widget_ids(&self) -> Vec<WidgetID> {
```
AaronM04: It might be good to have a `recursive` bool here.

conwayste/src/ui/ui_errors.rs
```diff
@@ -15,11 +15,33 @@
...
+use crate::ggez::GameError;
+
 custom_error! {pub UIError
     InvalidDimensions {reason: String} = "UIError::InvalidDimensions({reason})",
     WidgetNotFound {reason: String} = "UIError::WidgetNotFound({reason})",
     InvalidAction {reason: String} = "UIError::InvalidAction({reason})",
-    ActionRestricted{reason: String} = "UIError::ActionRestricted({reason})"
+    ActionRestricted{reason: String} = "UIError::ActionRestricted({reason})",
+    WidgetIDCollision{reason: String} = "UIError::WidgetIDCollision({reason})",
+    GgezError{reason: String} = "UIError::GgezError({reason})"
```
AaronM04: Not sure this is a good idea -- it is probably better to preserve the `GameError` rather than making a new string out of its string message.

conwayste/src/ui/widget.rs
```diff
@@ -68,6 +71,16 @@ pub trait Widget: Downcast {
         Ok(())
     }
 
+    /// Textfield gains focus and begins accepting user input
```
AaronM04: Should say widget here and below. There could be more than one kind of thing that takes focus.

conwayste/src/ui/common.rs
```diff
@@ -25,22 +25,18 @@ use crate::constants::DEFAULT_UI_FONT_SCALE;
...
                 match widget_result {
                     Ok(widget) => {
                         return widget.downcast_mut::<$type>();
                     }
                     Err(e) => {
-                        info!("Could not find $type widget of {:?} in layer of {:?}! {:?}",
-                        id,
-                        layer_id,
-                        e);
+                        info!("Could not find $type widget of {:?} in layer! {:?}", id, e);
```
AaronM04: I'm not really a fan of macros that add stuff like this (log lines, etc.). You can address here or add a TODO (since this was already there before your PR).

conwayste/src/ui/layer.rs
```diff
@@ -31,137 +31,251 @@ use super::{
...
-            tf.enter_focus();
-            self.focused_widget = Some(id);
-            return Ok(());
+    /// Retreives a mutable reference to a widget. This will search one Pane-level deep
+    /// for the provided widget-id.
+    ///
+    /// # Error
+    /// A WidgetNotFound error will be returned if the widget-id is not found.
+    /// does not exist in the internal list of widgets.
+    pub fn get_widget_mut(&mut self, widget_id: WidgetID) -> UIResult<&mut Box<dyn Widget>> {
```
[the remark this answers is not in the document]
Manghi: Thoughts on using something like this?
```rust
type BoxedWidget = Box<dyn Widget>;
```
and then we can do
```rust
 pub fn get_widget_mut(&mut self, widget_id: WidgetID) -> UIResult<&mut BoxedWidget>
```

conwayste/src/ui/layer.rs
```diff
@@ -31,137 +31,251 @@ use super::{
...
 }
 
-/// A container of one or more widgets or panes
-impl Layer {
-    /// Specify the unique widget identifer for the layer
-    pub fn new(widget_id: WidgetID) -> Self {
-        Layer {
-            id: widget_id,
-            widgets: vec![],
+/// A container of one or more widgets or panes ordered by virtual layers.
```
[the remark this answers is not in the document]
Manghi: Because we're using a number to describe the z-order, there isn't a grouping/collection for widgets at a depth (unlike before). Thus I described it as virtual here. But I think I'll just replace it with something else. Let me know if the later commit to address this is insufficient.

conwayste/src/ui/ui_errors.rs
Manghi: Yeah I agree with you now that you've pointed this out. I've created #87 to track this.

conwayste/src/ui/ui_errors.rs
Manghi: It was actually unused. Gunk from an attempt at converting between `UIError` and `GameError`s, which was resolved by the code written further below. Good catch!

conwayste/src/ui/common.rs
Manghi: Agreed it should be an error.

conwayste/src/ui/pane.rs
Manghi: Are you worried about stack usage? Not following how that would be better aside from that.

conwayste/src/ui/layer.rs
Manghi: It's a bug! Great catch. Should be `==`.

conwayste/src/ui/pane.rs
AaronM04: No, more about the caller not always wanting the widget IDs recursively.

conwayste/src/ui/layer.rs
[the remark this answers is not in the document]
AaronM04: I was thinking something like `&mut dyn Widget` so there is only one level of indirection.

See for example:
```rust
trait T {
    fn some_method(&mut self) -> i64;
}

#[derive(Default)]
struct S {
    pub x: i64,
}

impl T for S {
    fn some_method(&mut self) -> i64 {
        self.x += 1;
        self.x
    }
}

fn gets_a_t(tref: &mut dyn T) {
    tref.some_method();
}

fn main() {
    let mut box_t: Box<dyn T> = Box::new(S::default());
    gets_a_t(&mut *box_t);
}
```

conwayste/src/ui/pane.rs
Manghi: Can we add it in if we see the use case later down the road?

conwayste/src/ui/layer.rs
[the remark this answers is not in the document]
Manghi: That might be a bit tricky given that it's being returned, and we don't know the size of them (hence the `Box` and not `Sized`) for all `Widgets`. I don't see a clear path without trying out some things at least. 

Could we address this in a separate issue since this PR is nearly done?

conwayste/src/client.rs
```diff
@@ -386,7 +386,7 @@ impl MainState {
             GameError::ConfigError(msg)
         })?;
 
-        let ui_layout = UILayout::new(ctx, &config, font.clone());
+        let ui_layout = UILayout::new(ctx, &config, font.clone()).unwrap();
```
AaronM04: Please either fix the `.unwrap()` so this can't panic or explain why it can't happen in a comment. :)

conwayste/src/client.rs
```diff
@@ -626,10 +626,8 @@ impl EventHandler for MainState {
             Screen::Exit => {}
         }
 
-        if let Some(ref mut layers) = LayoutManager::get_screen_layers(&mut self.ui_layout, current_screen) {
-            for layer in layers.iter_mut() {
-                layer.draw(ctx)?;
-            }
+        if let Some(ref mut layering) = LayoutManager::get_screen_layering(&mut self.ui_layout, current_screen) {
```
AaronM04: Do you need the `ref` here?

conwayste/src/ui/pane.rs
AaronM04: Sure, ok.

conwayste/src/ui/layer.rs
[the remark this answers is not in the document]
AaronM04: OK

conwayste/src/ui/layer.rs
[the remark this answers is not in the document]
AaronM04: Can you explain how the optional transparency works? It's still... unclear... :sunglasses: . Also this would be a good place to say if the widget with the smaller or larger z-index is on top.

AaronM04:
@Manghi please resolve conflicts.

conwayste/src/ui/layer.rs
```diff
@@ -31,137 +32,245 @@ use super::{
...
+        self.rebuild_id_cache();
+        Ok(())
+    }
+
+    /// Removes a widget belonging to the layering
+    ///
+    /// # Error
+    /// A WidgetNotFound error can be returned if a widget with the `widget_id` does not exist
+    /// in the internal list of widgets.
+    pub fn _remove_widget(&mut self, widget_id: WidgetID) -> UIResult<()> {
```
AaronM04: Why does this have an underscore at the front?

conwayste/src/ui/layer.rs
```diff
@@ -31,137 +32,245 @@ use super::{
...
 use crate::constants::colors::*;
 
-pub struct Layer {
-    pub id: WidgetID,
-    pub widgets: Vec<Box<dyn Widget>>,
+pub struct Layering {
     pub with_transparency: bool,
-    pub focused_widget: Option<WidgetID>,
+    pub widget_list: Vec<BoxedWidget>,
+    focused_ids: Vec<Option<WidgetID>>,
```
AaronM04: It appears that the index is a z-index. Why is that the case? And Why are the elements `Option<WidgetID>`?

conwayste/src/ui/layer.rs
```diff
@@ -31,137 +32,245 @@ use super::{
...
+        self.id_cache.iter().find(|&&id| id == widget_id).is_some()
     }
 
-    /// Layer passes `exit_focus` request forward to its elements
-    pub fn exit_focus(&mut self) {
-        if let Some(other_id) = self.focused_widget {
-            if let Some(other_tf) = TextField::widget_from_id(self, other_id)
-            {
-                other_tf.exit_focus();
+    /// Returns an optional pair of indices if the widget-id is found in Pane beloning to
```
AaronM04: typo - `beloning` :)

conwayste/src/ui/layer.rs
```diff
@@ -31,137 +32,245 @@ use super::{
...
+            widget.on_drag(original_pos, current_pos);
         }
     }
 
-    fn draw(&mut self, ctx: &mut Context) -> GameResult<()> {
-        if self.with_transparency {
+    pub fn draw(&mut self, ctx: &mut Context) -> UIResult<()> {
+        let highest_z_index = self.peek_z_index();
+
+        if self.with_transparency && highest_z_index != 0 {
```
AaronM04: Please explain why this test is the way it is.

conwayste/src/ui/layer.rs
```diff
@@ -31,137 +32,245 @@ use super::{
...
-/// A container of one or more widgets or panes
-impl Layer {
-    /// Specify the unique widget identifer for the layer
-    pub fn new(widget_id: WidgetID) -> Self {
-        Layer {
-            id: widget_id,
-            widgets: vec![],
+/// A Layering is a container of one or more widgets or panes, where widgets are ordered by
+/// by their z_index to create the appearance of a layer, for a given game screen. Layerings
+/// support an optional transparency between two adjacent z-orders.
```
AaronM04: Is there one `Layering` per screen? Might want to say whether or not that's the case.

conwayste/src/ui/layer.rs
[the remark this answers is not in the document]
Manghi: Yes, thanks for this! Found a discrepancy with how I *thought* it worked versus what the code was doing. Updated in a3537c4.

conwayste/src/ui/layer.rs
Manghi: Not currently used. Clarified in 4f3d0c5.

conwayste/src/ui/layer.rs
Manghi: Found a behavioral bug with it due to your advise for clarifying the `Layering` comment. Bravo!

conwayste/src/ui/layer.rs
Manghi: `focused_ids` contains the widget-id with active focus for each z-order. If it is `None`, then no widget in the "layer" holds focus. The reason it's done by z-order is so that widget focus is retained for a z-order `n` after a new order `n+1` is created and then removed.

conwayste/src/client.rs
Manghi: Nope. Not the `mut` either. Fixed.

AaronM04:
I think maybe the semi-transparent layer should be determined by a method or property of the Pane -- if it's `true` *and* the Pane is focused, then draw the semi-transparent layer under it.

AaronM04:
It does not compile. Not sure if you knew that already :)

Manghi:
Yep, those were the unit tests that weren't updated yet.

Since your last review, the underlying data-structure changed to a tree, which simplified some things overall. 

Ready for your review whenever you get a chance. :smile:

conwayste/src/ui/pane.rs
```diff
@@ -145,32 +133,16 @@ impl Widget for Pane {
...
-                if within_widget(point, &w.rect()) {
-                    let ui_action = w.on_click(point);
-                    if ui_action.is_some() {
-                        return ui_action;
-                    }
-                }
-            }
-        }
+    fn on_click(&mut self, _point: &Point2<f32>) -> Option<(WidgetID, UIAction)> {
+        // TODO: Need to pass children nodes
```
AaronM04: I saw you removed the widgets Vec. Is the idea here that the Layering owns the widgets inside a Pane and passes them into Pane as needed? In my UIContext branch I was assuming Panes would own their widgets still.

AaronM04: LOL, I just read this in the `id_tree` crate description:
> In this implementation, the Tree owns all of the Nodes and all inter-Node relationships are managed with NodeIds.

I guess that answers that :grinning:

Manghi:
@AaronM04 Ready for your review! Rip it apart heh.

There's still some additional effort to get #92 working, but that will need the groundwork provided by #94 as previously stated. Given that we're diverging fast, I'd like to get this into `master` so I can help merging `master` into your `uicontext` branch.

AaronM04:
Silly me, I didn't approve it. :astonished:

PR #88, opened 2019-11-15
Set player name in game

AaronM04:
Add support for an Options screen that allows you to set your player name. It should be saved to the config when leaving the Options screen.

- [x] add new "Options" button to existing main menu pane - transition to Options screen
- [x] add support for Options screen in uilayout.rs
- [x] gaining focus of a textfield when clicking within the boundaries of the textfield
- [x] losing focus of a textfield when clicking outside the boundaries of the textfield
- [x] when entering Options screen, load Options from config (only player name for now)
- [x] when exiting Options screen, save Options into config
- [x] in draw & update, add support for Options screen
- [x] make ticket for moving other settings from old Options to new Options so the old menu system can be removed (see #162)
- [x] unit tests pass
- [x] no new warnings


conwayste/src/ui/pane.rs
```diff
@@ -97,6 +104,50 @@ impl Pane {
...
     }
     */
+
+    /// Adds the vector of children to the parent, and shrink the parent to fit the widgets with
+    /// padding on all sides. The widgets will have the padding added to their x and y coordinates.
+    /// Therefore it is suggested that the upper left widget in `children` be at `(0, 0)`.
+    ///
+    /// # Errors
+    ///
+    /// It is an error if the parent does not have an `add` method for adding `Widget`s.
```
Manghi: When would this occur?

Manghi:
Just had a single question about a comment you added. All else looks great so far.

conwayste/src/ui/pane.rs
AaronM04: Never. It's left over from before this was a method of `Pane`. It just took a `Box<dyn Widget>` before. Good catch!

AaronM04:
Putting this on hold until #83 is merged so I can use the widget focus stuff.

Manghi:
The merge conflicts weren't that bad to resolve, but I think it's better if we just approach this PR from a fresh mindset. The good changes you made in d3df6e328753c10eb4269cbba31f5723be2ffc9e need to be redone. I can try to get this if you're busy elsewhere.

AaronM04:
Yes please! I think it'll be a little while before I can compile anything ggez-based.

On July 15, 2020 9:03:35 PM PDT, Manghi <notifications@github.com> wrote:
>The merge conflicts weren't that bad to resolve, but I think it's
>better if we just approach this PR from a fresh mindset. The good
>changes you made in d3df6e328753c10eb4269cbba31f5723be2ffc9e need to be
>redone. I can try to get this if you're busy elsewhere.
>
>-- 
>You are receiving this because you were assigned.
>Reply to this email directly or view it on GitHub:
>https://github.com/conwayste/conwayste/pull/88#issuecomment-659144576

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Manghi:
Brief summary of what's been implemented from https://github.com/conwayste/conwayste/pull/88/commits/f876861ee2c0dc2a3fabbb256379429293babfe1 to 7d4e355:

1. You can now click in the rectangle of a textfield and have it gain focus
2. You can press escape to lose focus of a focused textfield
3. Moved a lot of the "Screen::Run" context (key handlers, mouse handlers, universe/game state) into `gamearea.rs`
4. Updated `UIContext` to pass in the viewport and gamearea state so widgets can interact with them

Working on task 4 next (lose textfield focus when clicking outside its boundary if it has focus)

AaronM04:
Only one test to fix:
```
---- viewport::test::test_gridview_default_instantiation stdout ----

thread 'viewport::test::test_gridview_default_instantiation' panicked at 'assertion failed: `(left == right)`

  left: `120`,

 right: `128`', conwayste/src/viewport.rs:371:9
```

AaronM04:
OK to check these 4 boxes?
* when entering Options screen, load Options from config (only player name for now)
* when exiting Options screen, save Options into config
* in draw & update, add support for Options screen
* make ticket for moving other settings from old Options to new Options so the old menu system can be removed

Manghi:
> OK to check these 4 boxes?

Nope. Not been implemented. I got fixated on getting "advanced" focus support working so with that done now I can actually work on these remaining four tasks.

I'll fix the unit test next as well. :D

conwayste/src/ui/layer.rs
```diff
@@ -92,7 +94,7 @@ pub struct Layering {
     // will have children nodes which are the nested elements
     // (think Buttons) of the widget.
     removed_node_ids:      HashSet<NodeId>, // Set of all node-ids that have been removed from the Tree
-    highest_z_order:       usize,           // Number of layers allocated in the system + 1
+    pub highest_z_order:   usize,           // Number of layers allocated in the system + 1
```
AaronM04, line 97: This is kind of a red flag to me. What is mutating this?

AaronM04 (commented):
Nice work! There are some issues I can see though:
* There has been a regression: hitting space while the game is in run mode no longer pauses.
* Setting the player name doesn't save it to the config or change it in the chatbox.
* Typing enter when finished typing the player name causes the text to disappear.
* Pressing Esc when the checkbox is focused does not have the desired result (going to the main menu).

conwayste/src/ui/layer.rs
Manghi: No mutation is going on. That is used in client.rs line 519 to determine if the `Run` state is at the lowest layer. I would have normally used a getter but you like to make these public so I oped for that., But it sounds like not in this case?

conwayste/src/uilayout.rs
```diff
@@ -45,11 +45,44 @@ macro_rules! add_layering_support {
...
+                }
+                Err(Box::new(crate::ui::UIError::InvalidArgument {
+                    reason: format!("{:?} not found in UI Layout", screen),
+                }))
+            }
+        }
+    };
+}
+
+macro_rules! add_layering_support_with_nonmut {
```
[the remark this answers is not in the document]
Manghi: Meant to rename these as this was initially a test. Will fix.

Manghi:
> Nice work! There are some issues I can see though:
> 
>     * There has been a regression: hitting space while the game is in run mode no longer pauses.

Thanks! 
This has been fixed in the latest commit.

>     * Setting the player name doesn't save it to the config or change it in the chatbox.
> 
>     * Typing enter when finished typing the player name causes the text to disappear.

The second and third bullets have not been implemented yet; that's why the associated tasks are still unchecked. They will be worked on next.

 
>     * Pressing Esc when the checkbox is focused does not have the desired result (going to the main menu).
If a widget/entity contains focus, I would argue the desired result should be to lose focus as Escape is commonly used to cancel/abort event in many applications. I don't think it makes sense to -- in this case -- go to the previous menu while a widget still has focus.

Manghi:
@AaronM04 

I'm not entirely sure what is needed specifically by the checkbox in the original PR: `in draw & update, add support for Options screen`. All screens are updated and drawn by default using the `draw()` and `EventType::Update` events so I don't think there's any additional support needed.

But everything else is ready for your review!

AaronM04:
Seems like the arrow keys no longer work. Is that a known issue?

Manghi:
> Seems like the arrow keys no longer work. Is that a known issue?

Yep. It broke due to moving some of the logic into game area's `keypress_handler()` which did not update the snapshot of `game_area_state`. Bug has been fixed in latest commit.

Please continue your review whenever you get a chance.

conwayste/src/client.rs
```diff
@@ -834,6 +828,20 @@ struct UniDrawParams {
 }
 
 impl MainState {
+    fn get_gamearea(&mut self) -> GameAreaState {
```
AaronM04, line 831: Why not have this return a `Result` instead of calling `error!` here? I'd prefer to have the `error!` call closer to the top level.

conwayste/src/client.rs
```diff
@@ -534,6 +524,9 @@ impl EventHandler for MainState {
             }
         }
 
+        // Refresh the game area state after processing all events
+        game_area_state = self.get_gamearea();
```
AaronM04, line 528: Is this the line that actually fixes the issue?

AaronM04:
I just made those two comments. After addressing them, this is good to merge! 🕺

conwayste/src/client.rs
Manghi: Yep!

Manghi: Yep!

PR #94, opened 2019-12-10
UIContext

AaronM04:
This adds a UIContext so that widget logic can happen within the widget definition rather than in client.rs.

- [x] allow setting handlers
- [x] Layer and Pane propagate Click events to the appropriate children
- [x] Layer and Pane propagate all events to the appropriate children
- [x] switch and activate buttons and checkboxes with keyboard
- [x] chatbox should use keyboard event system
- [x] get screen transition buttons to work again
- [x] no UI actions; widget IDs shouldn't really matter for our code
- [x] tests pass (except the two network-related ones which we'll fix ASAP)
- [x] no new warnings


AaronM04:
@Manghi I think I'm going to just make this a proof of concept for now because we have two branches making large changes to our UI stuff already. Once #83 and #88 are merged into master, I'll merge master into this branch and do this All The Way(tm).

conwayste/src/ui/context.rs
```diff
@@ -0,0 +1,216 @@
...
+    pub config: &'a mut config::Config,
+}
+
+// TODO: move this elsewhere; it's in here to keep separate from other code (avoid merge conflicts)
+#[allow(unused)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub enum EventType {
+    Click,
+    KeyPress,
+    Move, // mouse move
```
Manghi: What about adding an variable to `Click` so we can specify what button resulted in the event?
Also what about renaming `Move` to `MouseMove` so its' clear what the event is related to.

conwayste/src/ui/context.rs
```diff
@@ -0,0 +1,216 @@
...
+///             ...
+///         }
+///     }
+/// }
+/// // top level of the module
+/// impl_emit_event!(MyWidget, self.handlers);
+/// ```
+#[macro_export]
+macro_rules! impl_emit_event {
+    ($widget_name:ty, self.$handler_field:ident) => {
```
Manghi: Huh cool, didn't know you could do `self.$...`

conwayste/src/ui/context.rs
```diff
@@ -0,0 +1,216 @@
...
+            }
+
+            /// Emit an event -- call all handlers for this event's type (as long as they return NotHandled)
+            fn emit(&mut self, event: &crate::ui::context::Event, uictx: &mut crate::ui::context::UIContext) -> Result<(), Box<dyn std::error::Error>> {
+                use crate::ui::context::Handled::*;
+                // HACK: prevent a borrow error when calling handlers
+                let mut handlers = self.$handler_field
+                    .take()
+                    .ok_or_else(|| -> Box<dyn std::error::Error> {
+                        format!("during .on({:?}, ...): a .emit call is in progress for {} widget",
```
Manghi: Is the error message accurate here? It seems to match the one in `on(...)`, but we're actually in `emit(...)`. Just double checking.

conwayste/src/ui/pane.rs
```diff
@@ -57,7 +62,28 @@ impl Pane {
             previous_pos: None,
             border: 1.0,
             bg_color: None,
-        }
+            handlers: Some(context::HandlerMap::new()),
+        };
+        // TODO: propagate events for other EventTypes
+        let handler: context::Handler = Box::new(|obj, uictx, evt| {
```
Manghi: I think this would also benefit from a macro since it's nearly identical to the one in `Layer`.

conwayste/src/ui/context.rs
AaronM04: Yes to `MouseMove`! As for adding data to `Click` (or any other enum here), I think this may not be the best place.

If you look at `Packet`, just about every variant has data. I wonder if it made the enum more difficult to use. Thoughts?

conwayste/src/ui/context.rs
AaronM04: In a macro it's just a token tree. As long as you balance ([{, anything goes.

conwayste/src/ui/context.rs
AaronM04: Good catch. I'll take a closer look. I think my logic at the time was that if the handler field is `None`, either a `.on` (register handler) or a `.emit` (call handler) was already in progress for this widget.

conwayste/src/ui/context.rs
Manghi: I agree that it makes it difficult to use. Suggesting optional fields in the `Event` struct for fields that do not apply to all event types.

conwayste/src/ui/pane.rs
AaronM04: Done! I did something a little more abstract (set a handler for each event type that's a mouse event). See `forward_mouse_events!` macro.

I plan to do something similar for keyboard (which would use the focused widget not the widget under the cursor).

AaronM04:
At this point (unless you have any more review feedback @Manghi ), I'm going to pause this until #83 is merged.

conwayste/src/ui/context.rs
```diff
@@ -93,9 +95,27 @@ pub enum EventType {
 #[derive(Debug, Clone)]
 pub struct Event {
     pub what: EventType,
-    pub point: Point2<f32>,
+    pub point: Point2<f32>, // Must not be None if this is a mouse event type
```
Manghi: Does the comment actually apply here? Seems confusing as there isn't an `Option`.

Manghi:
> At this point (unless you have any more review feedback @Manghi ), I'm going to pause this until #83 is merged, so I know how to distribute keyboard events based on focus.

Probably a good idea since 83 changes how the underlying data is structured. No additional comments on my part for this PR. Great job with the clean up so far. :+1:

conwayste/src/ui/pane.rs
```diff
@@ -77,24 +78,37 @@ impl Pane {
...
+                                if let Some(emittable_ref) = widget_ref.as_emit_event() {
+                                    emittable_ref.emit(evt, &mut subuictx)?;
+                                } else {
+                                    warn!("Widget at point of click ({:?}) does not implement EmitEvent", evt.point);
+                                }
+                            }
+                        }
+                        Ok(Handled::Handled)
+                    },
+                );
```
Manghi: Nitpick, but you don't need the comma on line 102.

If `within_widget()` on line 94 never evaluated to true, why do we  still return `Handled` here on line 102? I'm wondering if we should only return `Handled` if we emitted on something (line 97).

AaronM04: That comma was added by rustfmt. I could take it out but if we ever rustfmt our whole project it'll come back. :disappointed: 

And I like your idea. I'll fix this.

Manghi: Ah it's easier to read too. :clap:

conwayste/src/ui/checkbox.rs
```diff
@@ -96,21 +105,26 @@ impl Checkbox {
...
-    pub fn toggle_checkbox(&mut self) -> bool {
-        self.enabled = !self.enabled;
-        self.enabled
+    fn click_handler(obj: &mut dyn EmitEvent, _uictx: &mut UIContext, _evt: &Event) -> Result<Handled, Box<dyn Error>> {
+        let mut checkbox = obj.downcast_mut::<Checkbox>().unwrap();
+
+        // toggle
+        checkbox.enabled = !checkbox.enabled;
+
+        Ok(Handled::NotHandled)
```
Manghi: Why does the handler returned `NotHandled` after the checkbox has been toggled?

Manghi:
_Sorry this took so long to get another review from my end._ 

Phenomenal work!! The `FocusCycle` is an excellent idea!

Regarding your comment of `TODO: consider removing self.video_settings`, I fully support this. I think if it's possible to use the `config` then there's no need for the archaic approach of copying those settings out of the config to some other structure in memory. At least I can't think of a good reason right now. If necessary, then it can be considered on a case-by-case basis, but the default/preferred behavior would be to use the `config`.

Could you explain where an `OpenEnded` focus cycle would be used? I see the comment stating it would be used for a Pane, but I also am wondering why one wouldn't want it to be always circular. 

Final comment: This design cycles through the widgets based on the order in which they were added, if I'm reading this correctly. We may want to consider (and this is a post-1.0 topic) cycling through the widgets in X/Y coordinate order. That way the cycle focuses on widgets in an ordered manner regardless of the order in which they were added. I've used programs that focus jump all over the place, leaving me (the user) very annoyed as it's hard to keep track of where the next focus will be. One possibility, for example, is traversing the a container from the top-left corner towards the bottom-right.

Great work man!

AaronM04:
That's because more handlers may need to run after this one. If it returned Handled they wouldn't run.

On March 17, 2020 12:19:27 AM PDT, Manghi <notifications@github.com> wrote:
>Manghi commented on this pull request.
>
>
>
>> +    fn click_handler(obj: &mut dyn EmitEvent, _uictx: &mut
>UIContext, _evt: &Event) -> Result<Handled, Box<dyn Error>> {
>+        let mut checkbox = obj.downcast_mut::<Checkbox>().unwrap();
>+
>+        // toggle
>+        checkbox.enabled = !checkbox.enabled;
>+
>+        Ok(Handled::NotHandled)
>
>Why does the handler returned `NotHandled` after the checkbox has been
>toggled?
>
>-- 
>You are receiving this because you were assigned.
>Reply to this email directly or view it on GitHub:
>https://github.com/conwayste/conwayste/pull/94#pullrequestreview-375787234

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

AaronM04:
> We may want to consider (and this is a post-1.0 topic) cycling through the widgets in X/Y coordinate order. 
I like this idea but I think I should do it as part of this PR.

Right before I got pulled onto my Rust presentation, I had decided an OpenEnded FocusCycle in each Pane is more trouble than it's worth. So that will go away (once I get done with my presentation).

Thanks for the review :grin: I _can't wait_ to get back to this!

conwayste/src/ui/context.rs
```diff
@@ -0,0 +1,374 @@
...
+                            // call each handler for this event type, until a Handled is returned
+                            for hdlr in handler_vec {
+                                let handled = hdlr(self, uictx, &event)?;
+                                if handled == Handled {
+                                    break;
+                                }
+                            }
+                        }
+                    }
+                }
```
Manghi, line 355: Is it possible to get stuck in this loop for ever? The use case I'm thinking -- and this is based off a scan of your phenomenal changes these past few commits -- is if a widget continues to forward events to itself while `Unhandled` is returned due to an implementation bug. If so, what do you think about logging something, for monitoring purposes, to see if we loop some `n` number of times and still have events to process.

AaronM04: It's not possible to get stuck in this loop from only returning Unhandled. The inner loop finishes when all the handlers have been called.

Manghi: I wrongly phrased my original question. Is it possible for `events.len() != 0` if a poorly written implementation continues to forward events to itself in one or more of it's handlers?

AaronM04:
@Manghi ready for your final review! :rocket:

Manghi:
I've been following your changes along the way and they looks awesome! I'm going to play around with it tomorrow and then do a final general review.

conwayste/src/ui/context.rs
```diff
@@ -0,0 +1,586 @@
...
+use ggez::input::keyboard::KeyCode;
+use ggez::nalgebra::Point2;
+use id_tree::NodeId;
+
+use super::treeview::TreeView;
+use super::BoxedWidget;
+use crate::config;
+use crate::Screen;
+
+pub struct UIContext<'a> {
```
Manghi, line 38: Can you add a rust doc here to explain what a `UIContext` is?

conwayste/src/ui/button.rs
```diff
@@ -112,15 +123,46 @@ impl Button {
...
+                EventType::LoseFocus => button.focused = false,
+                _ => unimplemented!("this handler is only for gaining/losing focus"),
+            };
+            Ok(Handled::NotHandled) // allow other handlers for this event type to be activated
+        };
+        b.on(EventType::GainFocus, Box::new(focus_chg.clone())).unwrap(); // unwrap OK b/c not being called within handler
+        b.on(EventType::LoseFocus, Box::new(focus_chg)).unwrap(); // unwrap OK b/c not being called within handler
+
+        // setup handler to forward a space keyboard event to the click handler
+        let keypress = |obj: &mut dyn EmitEvent, uictx: &mut UIContext, event: &Event| -> Result<Handled, Box<dyn Error>> {
```
Manghi, line 151: Any reason why this handler (and `focus_chg`) are defined as closures unlike `mouse_move_handler()`? Checkbox seems to go the latter route for all of its handler definitions.

conwayste/src/ui/context.rs
```diff
@@ -0,0 +1,586 @@
...
+                self.child_events.len(),
+            );
+        }
+    }
+}
+
+/// The type of an event.
+#[allow(unused)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, IntoEnumIterator)]
+pub enum EventType {
```
Manghi, line 179: Doesn't look like this `#[allow(unused)]` is needed. Can you also confirm the remaining uses of this in this file are still necessary?

conwayste/src/ui/context.rs
```diff
@@ -0,0 +1,586 @@
...
+/// Describes a MouseMove event in relation to a Rect.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum MoveCross {
+    Enter,
+    Exit,
+    None,
+}
+
+#[derive(Debug, Clone)]
+pub struct Event {
```
Manghi, line 206: Rust doc please!

conwayste/src/ui/context.rs
Manghi: Bump!

conwayste/src/ui/pane.rs
```diff
@@ -16,56 +16,313 @@
...
+
+        // for each event type, define a handler of the appropriate type (mouse or keyboard)
+        for event_type in EventType::into_enum_iter() {
+            if event_type.is_mouse_event() {
+                let handler = |_obj: &mut dyn EmitEvent,
+                               uictx: &mut context::UIContext,
+                               evt: &context::Event|
+                 -> Result<Handled, Box<dyn Error>> {
+                    // let pane = obj.downcast_mut::<Pane>()?; // uncomment and rename _obj to obj above if we need a Pane
+
```
Manghi, line 93: Please remove.

conwayste/src/ui/pane.rs
```diff
@@ -16,56 +16,313 @@
...
+            handler_data: context::HandlerData::new(),
+        };
+
+        // for each event type, define a handler of the appropriate type (mouse or keyboard)
+        for event_type in EventType::into_enum_iter() {
+            if event_type.is_mouse_event() {
+                let handler = |_obj: &mut dyn EmitEvent,
+                               uictx: &mut context::UIContext,
+                               evt: &context::Event|
+                 -> Result<Handled, Box<dyn Error>> {
```
Manghi, line 91: Same here regarding the handler in the body versus the `key_press_handler` down below.

conwayste/src/ui/pane.rs
```diff
@@ -16,56 +16,313 @@
...
+                    }
+                    Ok(Handled::NotHandled)
+                };
+                pane.on(event_type, Box::new(handler)).unwrap(); // unwrap OK because we aren't calling from within a handler
+            } else if event_type.is_key_event() {
+                // unwrap OK because we aren't calling from within a handler
+                pane.on(event_type, Box::new(Pane::key_press_handler))
+                    .unwrap();
+            } else {
+                // nothing to do if this is not a key or a mouse event
```
Manghi, line 123: Should we log this?

conwayste/src/ui/textfield.rs
```diff
@@ -92,21 +95,56 @@ impl TextField {
...
             visible_start_index: 0,
             font_info,
             bg_color: None,
+            handler_data: HandlerData::new(),
+        };
+        tf.on(EventType::KeyPress, Box::new(TextField::key_handler)).unwrap(); // unwrap OK b/c not inside handler now
+
+        // Set handlers for toggling has_keyboard_focus
+        let gain_focus_handler = move |obj: &mut dyn EmitEvent, _uictx: &mut UIContext, _evt: &Event|
+              -> Result<Handled, Box<dyn Error>> {
```
Manghi, line 116: In this file as well re: handlers.

conwayste/src/ui/treeview.rs
```diff
@@ -0,0 +1,309 @@
...
+            // zero out child
+            *child_ref = 0;
+        }
+        drop(view);
+        assert_eq!(*tree.get(&root_node_id).unwrap().data(), 1 + 2 + 3);
+    }
+
+    /*
+    #[test]
+    fn test_compile_fail_double_mut_borrow() {
```
Manghi, line 296: Will this test ever be needed?

conwayste/src/ui/ui_errors.rs
```diff
@@ -40,8 +40,11 @@ impl From<GameError> for Box<UIError> {
     }
 }
 
+/*
+ // This causes an infinite loop! FIXME: do we need this?
```
Manghi, line 44: Probably not. Much has to do be done with error handling (separately). Can just remove it for now?

conwayste/src/ui/button.rs
AaronM04: I started doing it this way just to show that it could be done. BTW a closure is the only nice way to capture shared state that can't be stored in the widget for whatever reason.

I can fix it.

conwayste/src/ui/context.rs
AaronM04: Done.

conwayste/src/ui/context.rs
AaronM04: Done.

conwayste/src/ui/ui_errors.rs
AaronM04: OK!

conwayste/src/ui/treeview.rs
AaronM04: Uncommenting it causes a compile failure -- if it doesn't, something got screwed up with borrowing.

I don't know of a way to have a test that verifies it can't compile, so I'm just leaving it commented.

AaronM04: BTW if you find a way to have a test with code that only passes if it fails to compile, that would be amazing! :grin:

conwayste/src/ui/pane.rs
AaronM04: I don't think so. If we need to add a handler to all Panes we can do it manually. No need to warn about events the Pane doesn't know how to register handlers for.

conwayste/src/ui/textfield.rs
AaronM04: Done.

AaronM04:
Github lost what the Bump comment was about.

conwayste/src/ui/pane.rs
AaronM04: Done.

AaronM04:
> Github lost what the Bump comment was about.

OK I think it's coming back to me -- you had concerns about the potential for infinite loops when handling forwarded events. I think we just have to be aware of that possibility when writing our handlers.

conwayste/src/ui/treeview.rs
Manghi: This looks promising: https://crates.io/crates/compiletest_rs
but I think it's overkill for this one test. Fine with leaving it as is.

PR #98, opened 2020-01-29
Use id_tree-generated NodeId's instead of statically defined WidgetID's

### LINKED ISSUE #97: Dynamically Assign Unique Widget Identifiers
Current scheme performs widget identifier definition at compile-time by hard-coding a unique `WidgetID(...)`, where the argument is an unsigned integer. While this was useful for initial UI creation and testing, this isn't maintainable for 1.0 as every widget requires a unique ID. This will easily grow to hundreds of widgets rapidly. Unique widget identifiers should be assigned at run-time without user intervention for ease of use and maintainability. 

Possible options thought up so far are:
* Reuse `NodeID` from `id-tree` crate
* Use an unique identifier generator library  like `snowflake`

First attempt will try to reuse `NodeID` from the `id-tree` crate

Manghi:
Resolves #97  

Simplified a lot of the code by using a `NodeId` instead of a `WidgetID`. 

Widgets are created without an identifier. As they are added to the layering's tree, `id_tree` generates a unique identifier which is then assigned to the widget. This is returned to the user in case they wish to keep track of a specific widget (like the chatbox, for example). 

Most of the time we use references to a `NodeId`, though there are a few exceptions to this rule and so a (relatively) cheap `clone` is used.

- [x] Try out option one to use `NodeId` from `id_tree` crate as a unique identifier
~~- [ ] Try out option two to use use our own implementation of `snowflake` or the like~~
- [x] Update copyright to 2020 for impacted files
- [x] Update unit tests
- [x] No new warnings
- [x] Verify no changes in behavior or functionality from prior implemenation

In the chatbox case, we'll eventually only need to keep track of the container's id instead of each component like is done in this PR. This will be much easier once #94 lands.

### DIFF
--- a/conwayste/src/client.rs
+++ b/conwayste/src/client.rs
@@ -79,7 +79,6 @@ use constants::{
     HISTORY_SIZE,
     INTRO_DURATION,
     INTRO_PAUSE_DURATION,
-    widget_ids::*,
     colors::*,
 };
 use input::{MouseAction, ScrollEvent};
@@ -91,7 +90,6 @@ use ui::{
     UIError,
     UIResult,
     Widget,
-    WidgetID,
 };
 use uilayout::UILayout;
 use uimanager::LayoutManager;
@@ -1439,47 +1437,23 @@ impl MainState {
         }
     }
 
-    fn handle_ui_action(&mut self, ctx: &mut Context, widget_id: WidgetID, action: UIAction) -> UIResult<()> {
-        match widget_id {
-            MAINMENU_PANE1_BUTTONYES
-            | MAINMENU_PANE1_BUTTONNO
-            | MAINMENU_TESTBUTTON  => {
-                match action {
-                    UIAction::ScreenTransition(s) => {
-                        self.screen_stack.push(s);
-                    }
-                    _ => {
-                        return Err(Box::new(UIError::InvalidAction{
-                            reason: format!("Widget: {:?}, Action: {:?}", widget_id, action)
-                        }));
-                    }
-                }
-            },
-            MAINMENU_TESTCHECKBOX => {
-                match action {
-                    UIAction::Toggle(enabled) => {
-                        self.config.modify(|settings| {
-                            settings.video.fullscreen = enabled;
-                        });
-                        self.video_settings.is_fullscreen = enabled;
-                        self.video_settings.update_fullscreen(ctx).unwrap(); // TODO: need ConwaysteError variant
-                    }
-                    _ => {
-                        return Err(Box::new(UIError::InvalidAction{
-                            reason: format!("Widget: {:?}, Action: {:?}", widget_id, action)
-                        }));
-                     }
-                }
-            },
-            MAINMENU_PANE1
-            | MAINMENU_LAYER1
-            | INGAME_LAYER1
-            | INGAME_PANE1 => {
-                return Err(Box::new(UIError::ActionRestricted{
-                    reason: format!("Widget: {:?} is either a Pane or Layer element. Actions are not allowed for Widgets of these types.", widget_id)
+    fn handle_ui_action(&mut self, ctx: &mut Context, action: UIAction) -> UIResult<()> {
+        match action {
+            UIAction::ScreenTransition(s) => {
+                self.screen_stack.push(s);
+            }
+            UIAction::Toggle(enabled) => {
+                self.config.modify(|settings| {
+                    settings.video.fullscreen = enabled;
+                });
+                self.video_settings.is_fullscreen = enabled;
+                self.video_settings.update_fullscreen(ctx).unwrap(); // TODO: need ConwaysteError variant
+            }
+            _ => {
+                return Err(Box::new(UIError::InvalidAction{
+                    reason: format!("Action: {:?}", action)
                 }));
-            },
-            ui::WidgetID(_) => {},
+            }
         }
 
         Ok(())
@@ -1488,8 +1462,9 @@ impl MainState {
     fn handle_user_chat_complete(&mut self, _ctx: &mut Context) {
         let username = self.config.get().user.name.clone();
         let mut msg = String::new();
+        let id = self.ui_layout.chatbox_tf_id.clone();
 
-        match TextField::widget_from_screen_and_id(&mut self.ui_layout, Screen::Run, INGAME_PANE1_CHATBOXTEXTFIELD) {
+        match TextField::widget_from_screen_and_id(&mut self.ui_layout, Screen::Run, &id) {
             Ok(tf) => {
                 if let Some(m) = tf.text() {
                     msg = format!("{}: {}", username, m);
@@ -1503,7 +1478,8 @@ impl MainState {
         }
 
         if !msg.is_empty() {
-            match Chatbox::widget_from_screen_and_id(&mut self.ui_layout, Screen::Run, INGAME_PANE1_CHATBOX) {
+            let id = self.ui_layout.chatbox_id.clone();
+            match Chatbox::widget_from_screen_and_id(&mut self.ui_layout, Screen::Run, &id) {
                 Ok(cb) => {
                     cb.add_message(msg.clone());
 
--- a/conwayste/src/client.rs
+++ b/conwayste/src/client.rs
@@ -485,8 +485,8 @@ impl EventHandler for MainState {
                     }
 
                     if left_mouse_click {
-                        if let Some( (ui_id, ui_action) ) = layer.on_click(&mouse_point) {
-                            self.handle_ui_action(ctx, ui_id, ui_action).or_else(|e| -> UIResult<()> {
+                        if let Some( ui_action ) = layer.on_click(&mouse_point) {
+                            self.handle_ui_action(ctx, ui_action).or_else(|e| -> UIResult<()> {
                                 error!("Failed to handle UI action: {}", e);
                                 Ok(())
                             }).unwrap();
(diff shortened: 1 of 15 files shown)


conwayste/src/ui/ui_errors.rs
```diff
@@ -22,7 +22,7 @@ custom_error! {pub UIError
     WidgetNotFound {reason: String} = "UIError::WidgetNotFound({reason})",
     InvalidAction {reason: String} = "UIError::InvalidAction({reason})",
     ActionRestricted{reason: String} = "UIError::ActionRestricted({reason})",
-    WidgetIDCollision{reason: String} = "UIError::WidgetIDCollision({reason})",
+    NodeIDCollision{reason: String} = "UIError::NodeIDCollision({reason})",
```
Manghi: I don't think this error, `NodeIDCollision` is physically possible due to the atomic uniqueness of `NodeID`?? I have no issues keeping it around and checking anyway if you think we should.

Manghi: Love that last idea!! I'd like to incorporate it in this PR. :smile: 

The case I thought of is if we somehow end up with a widget where a widget's node-id is `Some`, and it tries to be reinserted into the tree. Currently when we remove a widget from the tree, it's dropped. We don't return it to the caller. That case might only occur if we did return it to the caller. 

I think I'll leave the error in for now until the code matures a little more.

conwayste/src/ui/layer.rs
```diff
@@ -115,18 +128,12 @@ impl Layering {
         }
     }
 
-    /// Returns true if an entry with the provided WidgetID exists.
-    fn check_for_entry(&self, widget_id: WidgetID) -> bool {
-        // unwrap safe because a Layering should always have a dummy root-node
-        let root_id = self.widget_tree.root_node_id().unwrap();
+    /// Returns true if an entry with the provided NodeId exists.
+    fn check_for_entry(&self, id: &NodeId) -> bool {
```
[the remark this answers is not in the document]
Manghi: Sure will do. There isn't one present, but I will change this to a `get()` and check the `Result`. That should be quicker than a traversal. Thanks!

conwayste/src/client.rs
```diff
@@ -1070,11 +1067,12 @@ impl MainState {
                 return Ok(());
             }
             KeyCode::Return => {
-                match TextField::widget_from_screen_and_id(&mut self.ui_layout, Screen::Run, INGAME_PANE1_CHATBOXTEXTFIELD) {
+                let id = self.ui_layout.chatbox_tf_id.clone();
```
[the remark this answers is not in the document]
Manghi: Agreed. This was always intended to be a temporary workaround until all of the UI branches finalize.

conwayste/src/ui/layer.rs
```diff
@@ -176,111 +170,97 @@ impl Layering {
...
-        if self.check_for_entry(widget_id) {
-            return Err(Box::new(UIError::WidgetIDCollision {
-                reason: format!(
-                    "Widget with ID {:?} exists in layer's widget list.",
-                    widget_id
-                ),
-            }));
+    ) -> UIResult<NodeId> {
+        // Check that we aren't inserting a widget into the tree that already exists
+         if let Some(id) = widget.id() {
```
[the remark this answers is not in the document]
Manghi: Ah yes! I alluded to this above and didn't realize I wrote it in already. Will fix!

conwayste/src/client.rs
```diff
@@ -1488,8 +1463,9 @@ impl MainState {
     fn handle_user_chat_complete(&mut self, _ctx: &mut Context) {
         let username = self.config.get().user.name.clone();
         let mut msg = String::new();
+        let id = self.ui_layout.chatbox_tf_id.clone();
```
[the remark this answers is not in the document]
Manghi: I cannot borrow an immutable reference to it because `self.ui_layout` is borrowed mutably as well.

conwayste/src/client.rs
```diff
@@ -1503,7 +1479,8 @@ impl MainState {
         }
 
         if !msg.is_empty() {
-            match Chatbox::widget_from_screen_and_id(&mut self.ui_layout, Screen::Run, INGAME_PANE1_CHATBOX) {
+            let id = self.ui_layout.chatbox_id.clone();
```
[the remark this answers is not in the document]
Manghi: See above.

conwayste/src/ui/common.rs
```diff
@@ -28,7 +28,7 @@ macro_rules! widget_from_id {
         use super::layer::Layering;
 
         impl $type {
-            pub fn widget_from_id(layer: &mut Layering, id: WidgetID) -> UIResult<&mut $type>
+            pub fn widget_from_id<'a, 'b>(layer: &'b mut Layering, id: &'a NodeId) -> UIResult<&'b mut $type>
```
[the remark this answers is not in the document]
Manghi: If they share the same lifetime (including the return value).

```Rust
error[E0515]: cannot return value referencing local variable `id`
  --> conwayste/src/uimanager.rs:69:24
   |
69 |                 return TextField::widget_from_id(layer, &id);
   |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^---^
   |                        |                                |
   |                        |                                `id` is borrowed here
   |                        returns a value referencing data owned by the current function

error: aborting due to previous error
```

Manghi:
Thanks for the review! Comments address and resubmitted for second round.

conwayste/src/ui/layer.rs
```diff
@@ -183,12 +189,10 @@ impl Layering {
...
         // Check that we aren't inserting a widget into the tree that already exists
-         if let Some(id) = widget.id() {
-            if self.check_for_entry(id) {
-                return Err(Box::new(UIError::NodeIDCollision {
-                    reason: format!("Widget with ID {:?} exists in layer's widget tree.", id),
-                }));
-            }
+        if let Some(id) = widget.id() {
+            return Err(Box::new(UIError::NodeIDCollision {
+                reason: format!("Widget with ID {:?} exists was assigned an ID already.", id),
```
AaronM04: Not what I was expecting, but I like it. Carry on...

conwayste/src/ui/layer.rs
```diff
@@ -260,25 +264,42 @@ impl Layering {
         let node = self.widget_tree.get_mut(&inserted_node_id).unwrap();
         node.data_mut().set_id(inserted_node_id.clone());
 
+        // Report the error but keep going
+        if self.removed_node_ids.contains(&inserted_node_id) {
+            error!("NodeId {:?} found in HashSet during widget insertion. Possible reusage! Please report the bug!", inserted_node_id);
```
AaronM04: I'd prefer that this be `warn!`.

conwayste/src/ui/layer.rs
```diff
@@ -260,25 +264,42 @@ impl Layering {
...
+        // clone is okay because it is required
+        self.widget_tree.remove_node(id.clone(), RemoveBehavior::DropChildren).or_else(|e| {
+            return Err(Box::new(UIError::InvalidAction {
+                // clone is okay for error reporting
+                reason: format!("NodeIDError occurred during removal of {:?}: {:?}", id.clone(), e)
+            }));
+        });
+
+        // clone is okay because the HashSet is intended to keep track of all removed widgets
+        if !self.removed_node_ids.insert(id.clone()) {
```
AaronM04: Shouldn't the IDs of all nodes under this node be add as well? You need a `traverse_level_order_ids` here.

conwayste/src/ui/layer.rs
```diff
@@ -260,25 +264,42 @@ impl Layering {
...
+            error!("NodeId {:?} found in HashSet during widget removal. Possible reusage! Please report the bug!", id);
```
AaronM04: If it's already in the set, I say just ignore this. It already got logged when the widget was added.

conwayste/src/ui/layer.rs
```diff
@@ -696,4 +718,34 @@ mod test {
...
+            .add_widget(Box::new(pane), InsertLocation::AtCurrentLayer).unwrap();
+
+        let removal = layer_info.remove_widget(pane_id.clone());
+        assert_eq!(removal.is_ok(), true);
+
+        assert!(layer_info.removed_node_ids.contains(&pane_id));
+    }
+
+        #[test]
+        #[should_panic]
```
AaronM04: Please de-indent these two lines. Also, what is panicking here?

If it's an unwrap, I'd rather have `assert_eq!` to check that it's a certain error value.

conwayste/src/uilayout.rs
```diff
@@ -75,12 +75,8 @@ impl UILayout {
             chatbox_font_info,
             constants::CHATBOX_HISTORY
         );
-        match chatbox.set_rect(chatbox_rect) {
-            Ok(()) => { },
-            Err(e) => {
-                error!("Could not set size for chatbox during initialization! {:?}", e);
-            }
-        }
+        chatbox.set_rect(chatbox_rect)?;
```
AaronM04: Noice :)

conwayste/src/ui/layer.rs
Manghi: Whoops, thanks. I can see the confusion since there are multiple calls to `unwrap()`.

conwayste/src/ui/layer.rs
Manghi: Good catch! Will fix in next commit.

PR #129, opened 2020-06-14
Add Server Status Command

Manghi:
- [x] Add server status packet types (GetStatus/Status) to `net.rs`
- [x] Update netwayste client and server code to support server status query
- [x] Add client-side latency measurement in GetStatus/Status
- [x] Add server-side latency measurement in Update/UpdateReply
- [x] Add unit tests where applicable
- [x] No new warnings
- [x] Address `PR_GATE`s
- [x] Fix bug causing dissector to reach `unreachable!()` during an `get_ett_addr()` 
    Fixed in  9816ee9

*(Easier to view diff when using 27f008a as the baseline)*

### DIFF
--- a/netwayste/src/utils/ping.rs
+++ b/netwayste/src/utils/ping.rs
@@ -67,6 +67,13 @@ impl PingFilter {
     pub fn update(&mut self) {
         if !self.in_progress {
             error!("The PingFilter's start() was not called so a duration cannot be computed.");
+            let elapsed = self.start_timestamp.elapsed();
+            error!(
+                "PingFilter.start_timestamp snapshot was {}.{}.{} seconds ago.",
+                elapsed.as_secs(),
+                elapsed.as_millis(),
+                elapsed.as_micros()
+            );
         }
 
         let latency = Instant::now().duration_since(self.start_timestamp);
@@ -151,5 +158,4 @@ mod tests {
 
         assert_eq!(pf.average_latency_ms, Some(650));
     }
-
 }
--- a/netwayste/src/utils/ping.rs
+++ b/netwayste/src/utils/ping.rs
@@ -116,8 +116,6 @@ impl LatencyFilter {
             let average_rtt_ms = (self.running_sum as f64 / LATENCY_FILTER_DEPTH as f64) as u64;
             let average_latency_ms = average_rtt_ms / 2;
 
-            // TODO: Remove me once the server list is implemented in the UI
-            println!("Client-side Ping: {}", average_latency_ms);
             self.average_latency_ms = Some(average_latency_ms);
         }
 
--- a/netwayste/src/utils/ping.rs
+++ b/netwayste/src/utils/ping.rs
@@ -39,6 +39,7 @@ impl PingFilter {
         }
     }
 
+    #[deny(unused_variables)]
     pub fn reset(&mut self) {
         let Self {
             ref mut average_latency_ms,
--- a/netwayste/src/utils/ping.rs
+++ b/netwayste/src/utils/ping.rs
@@ -22,6 +22,7 @@ use std::time::Instant;
 /// meaningful average.
 const PING_FILTER_DEPTH: usize = 12;
 
+/// A moving-average filter used to level out the latencies calculated from network request/response times.
 pub struct PingFilter {
     pub average_latency_ms: Option<u64>,
     running_sum: u64,
@@ -65,7 +66,7 @@ impl PingFilter {
 
     pub fn update(&mut self) {
         if !self.in_progress {
-            panic!("start() was not called so a duration cannot be computed");
+            error!("The PingFilter's start() was not called so a duration cannot be computed.");
         }
 
         let latency = Instant::now().duration_since(self.start_timestamp);
@@ -151,10 +152,4 @@ mod tests {
         assert_eq!(pf.average_latency_ms, Some(650));
     }
 
-    #[test]
-    #[should_panic(expected = "start() was not called so a duration cannot be computed")]
-    fn test_ping_filter_update_called_before_start() {
-        let mut pf = PingFilter::new();
-        pf.update();
-    }
 }
--- a/netwayste/src/utils/ping.rs
+++ b/netwayste/src/utils/ping.rs
@@ -18,12 +18,14 @@
 use std::collections::VecDeque;
 use std::time::Instant;
 
+/// This indicates the number of samples needed by the ping filter to calculate a statistically
+/// meaningful average.
 const PING_FILTER_DEPTH: usize = 12;
 
 pub struct PingFilter {
-    pub average_latency_ms: u128,
-    running_sum: u128,
-    history: VecDeque<u128>,
+    pub average_latency_ms: Option<u64>,
+    running_sum: u64,
+    history: VecDeque<u64>,
     start_timestamp: Instant,
     in_progress: bool,
 }
@@ -31,7 +33,7 @@ pub struct PingFilter {
 impl PingFilter {
     pub fn new() -> PingFilter {
         PingFilter {
-            average_latency_ms: 0,
+            average_latency_ms: None,
             running_sum: 0,
             history: VecDeque::with_capacity(PING_FILTER_DEPTH),
             start_timestamp: Instant::now(),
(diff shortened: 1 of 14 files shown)


AaronM04:
Thanks for the PR! This is coming together :D

I have a question about the `PingFilter`. If the client sends out pings to multiple servers, how will this keep track of them? Do we need a `HashMap<String, PingFilter>` or will there be a map inside the filter?

netwayste/src/utils/ping.rs
```diff
@@ -0,0 +1,150 @@
...
+            average_latency_ms: 0,
+            running_sum: 0,
+            history: VecDeque::with_capacity(PING_FILTER_DEPTH),
+            start_timestamp: Instant::now(),
+            in_progress: false,
+        }
+    }
+
+    pub fn reset(&mut self) {
+        let Self {
```
AaronM04, line 43: You need a `#![deny(unused_variables)]` above this line for this pattern to be useful.

netwayste/src/utils/ping.rs
```diff
@@ -0,0 +1,150 @@
...
+ * this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+use std::collections::VecDeque;
+use std::time::Instant;
+
+const PING_FILTER_DEPTH: usize = 12;
+
+pub struct PingFilter {
+    pub average_latency_ms: u128,
```
AaronM04, line 24: Do we need `u128` here? I think this could be `u64`.
Also, it might be a good idea to wrap the number in an option, making the type be `Option<u64>`. That way, code that accesses this to get the ping time can easily tell the difference between it not being set yet and the ping time being `<1ms`.

netwayste/src/utils/ping.rs
```diff
@@ -0,0 +1,150 @@
...
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+use std::collections::VecDeque;
+use std::time::Instant;
+
+const PING_FILTER_DEPTH: usize = 12;
```
AaronM04, line 21: It would be nice to have a rustdoc comment indicating what this means.

netwayste/src/utils/ping.rs
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    pub fn start(&mut self) {
+        self.start_timestamp = Instant::now();
+        self.in_progress = true;
+    }
+
+    pub fn update(&mut self) {
+        if !self.in_progress {
+            panic!("start() was not called so a duration cannot be computed");
```
AaronM04, line 65: I would prefer to have an error returned since this code will probably be used by the registrar when we implement it, and it would suck to have it panic and stop working.

netwayste/src/utils/ping.rs
```diff
@@ -0,0 +1,150 @@
...
+
+        self.running_sum += latency_ms;
+        self.history.push_back(latency_ms);
+
+        // Wait for the filter to be populated
+        if self.history.len() > PING_FILTER_DEPTH {
+            // unwraps safe b/c of length check
+            let oldest = self.history.pop_front().unwrap();
+            self.running_sum -= oldest;
+            self.average_latency_ms = (self.running_sum as f64 / PING_FILTER_DEPTH as f64) as u128;
```
AaronM04, line 79: If you used `self.history.len()` as the denominator you could do this on every `update()` call rather than only when the history is full. :sunglasses:

AaronM04:
I think it would be a good idea to use the nonce in `PingFilter`, so for each nonce, you would need to keep track of the `Instant` at which it was sent. Then, for getting the latency, use the received nonce to look up that `Instant` and do the duration calculation with that.

AaronM04:
Can you explain your plans for these checkboxes a bit more?
>  Add server-side ping measurement
> More actions TBD...

netwayste/src/utils/ping.rs
Manghi: Ah yes thank you!

netwayste/src/utils/ping.rs
Manghi: `u128` is used here because that's the return type for a `Duration`. I had thought about trimming it down to `u64` since it's a safe cast. Looks like you agree so I'll do so!

Also good idea about the `Option`.

netwayste/src/utils/ping.rs
Manghi: I'll update this to an `error` for now, but please see a separate email for full response.

netwayste/src/utils/ping.rs
Manghi: See common below regarding a meaningful number of samples. It's not so much as being "full", but instead as having "enough".

Manghi:
> Thanks for the PR! This is coming together :D
> 
> I have a question about the `PingFilter`. If the client sends out pings to multiple servers, how will this keep track of them? Do we need a `HashMap<String, PingFilter>` or will there be a map inside the filter?

You're welcome, this was fun to work on!

The filter is only responsible for calculating a moving average of latencies. For the multiple server case, yes you'd could have something like a `HashMap<ServerId, PingFilter>` where `ServerId` is some unique identifier associated to the server. Each latency refresh (whether manually initiated via a "refresh" button, or automatic), produces a (statistically meaningful) average latency after `PING_FILTER_DEPTH` number of commands have been timed.

> I think it would be a good idea to use the nonce in PingFilter, so for each nonce, you would need to keep track of the Instant at which it was sent. Then, for getting the latency, use the received nonce to look up that Instant and do the duration calculation with that.

That would be useful for debugging when comparing debug dumps (for example) against a wireshark log. Sure! Aaaand now that I think about it, you wouldn't need the `in_progress` flag because it's directly associated to the nonce. 

> Can you explain your plans for these checkboxes a bit more?

Latency can be calculated from two viewpoints: either from the client as the initiator, or the server as the initiator. Currently the client piggy-backs off the `GetStatus/Status` while the server could piggy-back off of the `Update/UpdateReply`. In an ideal world these values are exactly the same so it'd be pointless, but I thought it might be useful early in development to know what the server's perspective is.

`More actions TBD` is just to address any additional changes that may arise during the review process (like using the nonce in the filter).

AaronM04:
For server-side ping, it might be a good idea to add an `Update::Ping{nonce:u64}` and `UpdateReply::Pong{nonce:u64}`. And I wouldn't say it would be pointless. It might be useful for changing the generation rate based on the ping times of the players.

And I would rather us not have to send a bunch of pings to get any results. In my opinion, it'd be better to have less accurate results if we can get them faster.

Manghi:
> And I would rather us not have to send a bunch of pings to get any results. In my opinion, it'd be better to have less accurate results if we can get them faster.

An accurately calculated latency is crucial. It is one of several key datum that gamers use to rate their game-play experience. An initial burst of a dozen pings over a few seconds won't slow things down.

Manghi:
All that's left is to fix a bug causing the dissector to reach `unreachable!()` during an `get_ett_addr()` call. I suspect it's related to adding of `GetStatus` / `Status`.

netwayste/src/utils/ping.rs
```diff
@@ -0,0 +1,189 @@
...
+        if self.history.len() > LATENCY_FILTER_DEPTH {
+            // unwraps safe b/c of length check
+            let oldest = self.history.pop_front().unwrap();
+            self.running_sum -= oldest;
+
+            // average round-trip time
+            let average_rtt_ms = (self.running_sum as f64 / LATENCY_FILTER_DEPTH as f64) as u64;
+            let average_latency_ms = average_rtt_ms / 2;
+
+            // TODO: Remove me once the server list is implemented in the UI
```
AaronM04, line 119: Why not `PR_GATE` this so we don't have a bunch of log output after the PR is merged?

AaronM04: If you want to see feedback maybe it would be better to have cli-client read the `average_latency_ms`.

Manghi: Check out the latest commit :smile: .

Also `PR_GATE`s are my way of keeping track things that must be resolved before the PR can be approve and merged.

Issue #108, opened 2020-05-13
Run rustfmt on entire workspace


Manghi:
Consider doing this on `master` after #107 and #94 are merged in. Would be nice to establish this as a merge requirement now that I got `rustfmt` working on my system.

I discovered we can have Travis fail if the code is not properly formatted. Might be over-aggressive? Other option is to just manually require it (like we commonly have for "no new warnings") after we run it on the entire workspace.

https://github.com/rust-lang/rustfmt#checking-style-on-a-ci-server

Manghi:
`netwayste` and `dissect-netwayste` have received this TLC.

AaronM04:
I think we should 1) get it set up so that it shows up as a failed check in PRs if any of the files are not rustfmted, then 2) have a single commit that rustfmts all .rs files in the repo. Ideally with a distinctive author name for visibility in git blame, etc.

AaronM04:
Useful link: https://crate-ci.github.io/pr/rustfmt.html

Also TIL that unlike gofmt, rustfmt is configurable! There's really only two options I think we should change from the default:
* max line length from 100 to 120, and
* struct_field_align_threshold to 20 (or at least pick a value between 10 and 25 or so)

Manghi:
>I think we should 1) get it set up so that it shows up as a failed check in PRs if any of the files are not rustfmted, then 2) have a single commit that rustfmts all .rs files in the repo. Ideally with a distinctive author name for visibility in git blame, etc.

Sounds good to me. Could we also specify the files/folders that were formatted? Something like: 
`[rustfmt] username: client.rs and all of conwayste/ui/`

> Also TIL that unlike gofmt, rustfmt is configurable! There's really only two options I think we should change from the default:

I'm fine with both of those configurations. 20 sounds good to me for `struct_field_align_threshold`.

I would also like to add:
* `empty_item_single_line` to false
* `enum_discrim_align_threshold` to the same as `struct_field_align_threshold`

PR #136, opened 2020-07-03
Rust-format all files, and make CI fail on PRs with any non-formatted files

AaronM04:
- [x] all .rs files formatted according to `.rustfmt.toml`
- [x] CI succeeds when all files are formatted (well, succeeds as far as it can...)
- [x] CI fails when one or more files are not formatted

Prerequisites on developer systems:

* Nightly rustfmt: `rustup +nightly component add rustfmt`

To format all files:

```cargo +nightly fmt -- -l```

To list which files are not formatted, but without actually formatting them:

```cargo +nightly fmt -- -l --check```


AaronM04:
Fixes #108

AaronM04:
@Manghi ready for your review!

This has one somewhat unrelated change in it to hopefully get Travis builds to pass again: https://github.com/conwayste/conwayste/pull/136/commits/118997f181de9972c186633d058fdf0aadfd0433

AaronM04:
The green checkmark is back!

There were two issues:
1) Compiling took too long (over 50 minutes, causing the build to be aborted by Travis) because we were running both with and without `--release`. Now it's only the release.
2) In some cases, the build would be aborted earlier because one step was running for more than 10 minutes without printing anything.

AaronM04:
Thanks @Manghi !

PR #139, opened 2020-07-15
Protocol updates

AaronM04:
These protocol changes should support the full game (1.0 at least). 😀


AaronM04:
I am labeling everything in net.rs `Packet` data model that needs implementing with `// TODO: add support`.

AaronM04:
@Manghi the branch is up to date with what we went over, plus I added one thing I forgot before:
`RequestAction::DropPattern`

Ready for review! 😀

AaronM04:
> One suggestion I have is to wrap `GenStateDiffPart::total_parts` around a type so that it, the mask, and the `part_number` can be managed internally by some `impl`. Reasoning is that we may wish to increase or decrease the count from 32 to something else and it should be easier to maintain if we need to tune that width. Let me know what you think.

Thanks for the review! I think we should have it be very explicit what data is in these packets. Also it is more time for us to support something like that, when it's unlikely we'd benefit from reducing it from 32 and increasing beyond 32 would mean we are sending at least 16 KB per generation (minimum MTU we're likely to encounter multiplied by 32), which means we should switch to a better diff encoding scheme. I'd rather use that time to get us closer to 1.0. 😁

PR #145, opened 2020-08-19
OpenBSD support in ggez

AaronM04:
This PR gets the GUI (ggez) client working on OpenBSD!

Caveats to note:

* This uses 4 git submodules (sorry) only until these changes can be pushed upstream. As the upstream repositories merge my code and publish it, we'll be able to switch to those versions in Cargo.toml instead of using `path =`, and then delete the corresponding submodules.

* gilrs (dependency of ggez) will never detect gamepads. I think this is ok for upstreaming... maybe.

* cpal (dependency of rodio, which in turn is a dependency of ggez) will cause a panic if an attempt is made to play sound. The support will probably have to be implemented before they accept a PR for this.

* It needs an environment variable `export LIBCLANG_PATH=/usr/local/lib` and package installed with `doas pkg_add llvm`. This is partially described here: https://github.com/mjkillough/sioctl-rs

@manghi can you please verify this works on a Linux system before approving?

```
git submodule update --init --recursive
cargo run --bin client
```


AaronM04:
Well Travis is passing, so that's something.

Remaining stuff:
* try to rebase my ggez commits onto upstream (not sure which of the 3 updated branches to use though)
* fix those unwraps in ggez related to rodio (may be fixed by the above)
* [avoid panic in gilrs when platform not supported by fixing ggez; remove gilrs submod too](https://gitlab.com/gilrs-project/gilrs/-/merge_requests/48#note_406599160)
* try to get cpal to actually work
* upstream all of this :)

Manghi:
Still builds and runs in Linux :+1:

AaronM04:
@Manghi I went ahead and merged this (since you said I could go ahead a few sessions back) 😀 

Once you pull master and merge it into your branche(s), you'll want to run this:
```
git submodule init && git submodule update
```

Let me know if you have any issues with these changes!

PR #152, opened 2020-09-11
Net: fix time in ping filter error message

AaronM04:
This is not a big deal but I just wanted to make a PR for this bug that jumped out at me. 😀 

Ready for review!


Manghi (approved):
Ah great catch! My only suggestion would be to keep the division of `seconds.millis.micro` as it's a little easier to digest with the period as the delimiter.

AaronM04:
I would much rather have just a single decimal point. How about just milliseconds and no more digits?

Manghi:
Let's keep it the way it currently is in the PR then. I prefer the higher resolution timestamp.

OK2Merge!

AaronM04:
TYVM!

PR #78, opened 2019-10-22
Load RLE pattern on client start-up, and make universe in client mode

AaronM04:
Currently this loses the fog.

PR #163, opened 2020-10-11
Changed uni pattern init to use RLE description;

Manghi:
Direct clone of #78 

Did not rebase b/c that branch is soooooo old. Used `player_name` as base as that has the latest and greatest client/universe changes, and its easier to see what this PR actually contains. We can merge that in first, and then I'll change the target of the PR to master.

* Changed uni pattern init to use RLE description;
* Universe also started in client mode for single-player experience


AaronM04:
Nice work! You deleted a lot of superfluous lines :) 

However I see one issue:
* No fog. Perhaps `force_known` is not the right approach here.

Issue #60, opened 2019-09-08
Implement server support for registrar


Manghi:
#31

PR #199, opened 2021-03-17
Server registrar support

AaronM04:
This PR adds support for registering the current server with the registrar. Since we don't have an Official Conwayste Registrar running yet, you'll have to run your own locally (see https://github.com/conwayste/registrar) to test this.

Once the registrar is running, try something like this:

```
cargo run --bin server --public-address 127.0.0.1:2016 --registrar-url http://127.0.0.1/addServer
```

And then view it in the registrar by going to http://127.0.0.1:8000/servers

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,8 @@ The Conwayste client and server compile and run right out of the box. Skip direc
 
 On Linux, the ALSA development files are required. These are provided as part of the `libasound2-dev` package on Debian and Ubuntu distributions and `alsa-lib-devel` on Fedora. For any other distribution, please refer to your package manager and/or compile them from source.
 
+Also install OpenSSL development files with `openssl-dev` or `openssl-devel` (depending on your distro).
+
 ## OpenBSD
 
 ```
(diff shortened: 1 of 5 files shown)


AaronM04:
Fixes #60

AaronM04:
@Manghi this is ready for your review! If you approve, please also merge as usual :)

Also please review the main branch of https://github.com/conwayste/registrar. At this point, everything is there except paging but that can wait until post-1.0, I think. I have a TODO. Let me know if you want me to create a GitHub issue for it.

The latest version of the registrar is running on https://registry.conwayste.rs/servers.

README.md
```diff
@@ -55,6 +57,31 @@ You will also need this environment variable. Add to your profile if desired:
...
+```
+
+See https://github.com/sfackler/rust-openssl/issues/1278#issuecomment-678597781
+
+Alternatively, change the line starting with `reqwest` in `netwayste/Cargo.toml` to have `default-features = false` with an extra `"rustls"` feature in the `features` array. Something like this:
+
+```
+reqwest = { version = "0.11", default-features = false, features = ["json", "rustls"] }
+```
+
```
Manghi, line 84: Can we instead include a static library of openssl in our releases so the user doesn't have to do this? This is a lot of work for anyone who just wants to `git clone` or `cargo install` the game.

Manghi (commented):
Everything looks good. This includes my review of the Go registrar. Let me know what you think about the one question I had related to the ssl library.

README.md
AaronM04: That is possible but I'm not sure it would be practical for a two developer project like ours. I think we should try using an all-Rust replacement like [rustls](https://crates.io/crates/rustls) if the current solution creates a bad user experience.

Manghi: Sounds like a great idea! We can try that out another day, merge will proceed.

Manghi (approved):

Issue #153, opened 2020-09-11
Filter Layer (New Netwayste Architecture)


AaronM04:
*Ticket 2*
Filter Layer (middle layer)

```
 <---To App----   +----------------------------+  ----To Transport--->
    (above)       |                            |       (below)
                  |                            |
      Filter      |                            |
     Command      |                            |   Transport             
     Requests     |                            |   Command       
   +-------------->                            |   Requests     
                  |                            +--------------->
      Filter      |         Filter             |                
     Responses    |                            |  Transport 
   <--------------+                            | Responses    
  & Notifications |                            <----------------
                  |                            |                
                  |                            |  Notifications             
                  |                            <--------------  
                  +----------------------------+ 
```                                                             

The filter layer is the middle layer sitting in between the application and transport layers. It contains five channels (3 facing the transport layer, 2 facing the application layer).

Responsibilities
----------------

1. It knows the meanings of packets (requests, updates, etc.), but it still does not know if the endpoint is a client or server.
2. For `RequestAction` and `ResponseCodes`, it handles reordering, removing duplicates and out-dated packets. 
3. For `Update` and `UpdateReply`, it handles fragmentation and reassembly. Filter knows about confirmed generations for each player (via Update/UpdateReply)
4. Manages the ping filter
5. DNS resolution; stores the following mapping: `{ (String , Port) : SocketAddr}`.  String is either an "IP" or "Hostname". 
    a. Keys are used by the app layer.
    b. Values are used by the transport layer. "Given" to transport layer during endpoint creation.
6. Authentication; 
   a. *Client operation*: If it receives a cookie in a ResponseCode, it
      associates the cookie with that endpoint it received it from. Any Packets sent
      to that endpoint going forward will use that cookie. 
   b. *Server operation*: if it receives a notification of a received packet
      from the Transport layer and it's never seen that endpoint before, it sends a
      Notification to the app layer above it. If the App layer then sends an
      Authenticated command, the filter layer creates a new cookie, associates it
      with that endpoint, and sends it back in a ResponseCode. It drops non-Connect
      packets from that endpoint without that cookie.
7. Keeps track of chat and game update sequences (see `Packet::Update` and `Packet::UpdateReply`)

Filter initialization will involve specifying the packet types (as a list) that it can accept. (For example, if used for a server, its initializer would tell it to not accept ResponseCodes).

Filter-Supported Commands (from app to filter)
----------------------------------------------

1. `SendRequestAction(endpoint, RequestAction)`
[...]

Issue #151, opened 2020-09-07
Transport Layer (New Netwayste Architecture)


AaronM04:
Transport Layer (lowest layer)

```
                  +----------------------------+
     Command      |                            |
     Requests     |                            |
   +-------------->                            |
                  |                            | TX to UDP
     Responses    |         Transport          +------------------>
   <--------------+                            |
                  |                            +<------------------+
                  |                            | RX from UDP
     Notifications|                            |
   <--------------+                            |
                  +----------------------------+
```

The transport layer is the lowest layer and contains three channels for communicating with the filter layer immediately above it. The three channels are:
1. Inbound channel that the filter layer uses to send commands to the transport layer.
2. Outbound channel that the transport layer uses to reply to a processed command
3. Outbound channel for notifications. The first notification will be notify the filter that packets have been received and are ready to be requested.

Responsibilities:
1. Handle sending or resending tx queue packets on the wire, per endpoint.
2. Buffers packets into individual rx queues for every endpoint. 

Endpoints: 
`SocketAddr` (includes address and port)
No knowledge about whether the endpoint is a client or some other server.

The transport layer does not know about the contents or meanings of any packets. 

Transport Supported Commands:
1. Get tx/rx packet queue count (Args: TX or RX queue, endpoint)
2. Take rx queue packets (Args: number of packets, endpoint, index in queue to consume) 
3. Send packets (Args: endpoint, vec<Packet Info (see below)>)
4. Drop endpoint (Arg: endpoint); Will drop all data associated with the endpoint
5. Cancel tx packet (Args: endpoint, vec<indexes to cancel>)
6. Create new endpoint (Args: endpoint, timeout (max Duration allowed between received packets))

Transport-supported Command Responses:
1. `BufferFull`
2. `NoPacketAtIndex`
3. `Accepted` (command is OK)
4. `QueueCount(num)`
5. `TakenPacket(packet)`
6. `ExceedsMtu`
7. `EndpointTimeout(endpoint)`

Packet Info:
Contains the `Packet`, retry count (u16), retry interval (usize).

Common features of all layers
---------------
* The `new` methods should take as one of the arguments a `TimeFns` struct with
  a) a method that returns the current `std::time::Instant` and b) a method
that, when given a `std::time::Duration`, returns a `Future` that completes
when that duration has elapsed. The rationale here is to allow easy unit
testing.

AaronM04:
@Manghi one down, 3 more to go (filter, client app, server app) 😁 🚀

[...]

PR #200, opened 2021-04-16
Netwayste Version 2

Manghi:
Changes for the Netwayste Re-architecture as related to #151 and #153 .

Goal is to make the Conwayste networking code easier to maintain and grow by reducing the coupling between modules, features, and implementation.


AaronM04:
I feel really good about this branch! :D

Mind if I push changes or would you prefer that I just review as you work on it?

Manghi:
> I feel really good about this branch! :D
> 
> Mind if I push changes or would you prefer that I just review as you work on it?

@AaronM04 
Commit and push as you'd like!

netwaystev2/src/transport/transport.rs
```diff
@@ -0,0 +1,231 @@
...
+                udp_stream_recv,
+                endpoints: EndpointData::new(),
+            },
+            cmd_tx,
+            rsp_rx,
+            notice_rx,
+        ))
+    }
+
+    pub async fn monitor(&mut self) -> Result<()> {
```
AaronM04, line 82: Can we name this something else, like `run`, since it's doing more than just monitoring. I would expect something that "monitors" to be a read-only thing...

netwaystev2/src/transport/transport.rs
```diff
@@ -21,14 +21,14 @@ use tokio::time as TokioTime;
...
 
-type TransportCmdSend = Sender<TransportCmd>;
+pub type TransportCmdSend = Sender<TransportCmd>;
 type TransportCmdRecv = Receiver<TransportCmd>;
 
 type TransportRspSend = Sender<TransportRsp>;
-type TransportRspRecv = Receiver<TransportRsp>;
+pub type TransportRspRecv = Receiver<TransportRsp>;
 
 type TransportNotifySend = Sender<TransportNotice>;
```
AaronM04, line 30: Should these be `pub` as well?

Manghi: Possibly. Right now it's only those that cross the Filter/Transport bridge.

netwaystev2/src/filter/sortedbuffer.rs
```diff
@@ -0,0 +1,251 @@
...
+        }
+        false
+    }
+
+    pub fn count_contiguous(&self, mut sequence: u64) -> usize {
+        let mut count = 0;
+        for tuple in &self.heap {
+            if tuple.0.0 == sequence {
+                count += 1;
+                sequence += 1;
```
AaronM04, line 31: Will this wrap correctly without panicking? @Manghi

netwaystev2/src/filter/sortedbuffer.rs
```diff
@@ -0,0 +1,251 @@
...
+        Some(self.cmp(other))
+    }
+}
+
+// Marker trait impl for Ord without needing to pass `Eq` down to the item
+impl<T> Eq for SequencedItem<T> {
+}
+
+#[cfg(test)]
+mod test {
```
AaronM04, line 91: What do you think about moving all these tests to a separate file, and in general, keeping our tests separate from our business logic?

netwaystev2/src/filter/sortedbuffer.rs
AaronM04: Also, I noticed the tests are failing for this because there is no way to iterate over the heap in the order that `pop()` will happen (without consuming the heap, which would be inefficient). I wonder if we really need this operation since we may not be able to implement it in an efficient way?

netwaystev2/src/filter/sortedbuffer.rs
Manghi: Love the idea. We tend to write quite a bit of tests anyways which makes the files unmanageable during navigation.

Manghi: Needed to make one function public in order to access it. I tried `pub(crate)` first but that didn't work since tests are seen as external to the crate.

netwaystev2/src/filter/sortedbuffer.rs
Manghi: > without consuming the heap, which would be inefficient
Exactly!

Thinking some more about this, you're absolutely right, we don't *really* need the operation. There aren't any great sorted set libs presently available for our use case anyways.

I think comparing against the `sequence` via `peek()` before a `pop()` in a loop is the right way to go for now. This is corrected in the latest commit.

netwaystev2/src/filter/sortedbuffer.rs
AaronM04: Well... they don't have to be _that_ separate... you have them as integration tests: https://doc.rust-lang.org/book/ch11-03-test-organization.html#integration-tests

That said, it's not a big deal that `contains_sequence_number` is public now.

netwaystev2/src/filter/interface.rs
```diff
@@ -0,0 +1,242 @@
...
+
+#[derive(Serialize, Deserialize, PartialEq, Debug, Clone)]
+pub struct RoomList {
+    pub room_name:    String,
+    pub player_count: u8,
+    // TODO: add support
+    pub in_progress:  bool,
+}
+
+pub enum FilterEndpointData {
```
AaronM04, line 214: What do you think about putting this somewhere else? I feel like this and the error below it don't belong in `interface.rs` since they are internal to the Filter.

Manghi: Yep agreed! There's a bunch of stuff here that doesn't belong -- rather, a lot of it is tailed towards the app-layer. I'll move it out.

netwaystev2/src/filter/sortedbuffer.rs
Manghi: I was thinking about this this morning and decided to go with `src/filter/tests` instead. Will be in latest set of commits once I push.

netwaystev2/src/filter/interface.rs
Manghi: I created a `src/protocol` folder which contains the data structure definitions for `Packet` and the like. Let me know what you think!

Manghi: Wasn't sure if I should move `PingPong` too but I decided to leave it in `filter.rs` for now.

AaronM04: I really like the organization in `src/protocol`. I can actually find structs and enums with this.

The only thing that I would change is this:

```rust
        /* PR_GATE add in later
        game_updates:    Vec<GameUpdate>, // Information pertaining to a game tick update.
        universe_update: UniUpdate,       // TODO: add support
        */
```
Can you make the complete structure we have in the `netwayste` crate be here too (no temporarily commented data structures) please?

Also I am not sure why we need a `Default` implementation for `Packet...?

AaronM04: The other thought I had is maybe we could merge `response.rs` into `request.rs`. Otherwise there will need to be an `updatereply.rs` file added to be consistent, and that seems like a lot of files. 😁

Manghi: Removed the `PR_GATE` comments and added in their associated data structures. 

I like that `request.rs` and `response.rs` are separate. The idea was that any data structures used within a `Packet` variant's would be defined in their respective files for quick look up. `UpdateReply` uses only primitive data types so there wasn't a need to define an `updatereply.rs`.... yet.

AaronM04:
It turns out we forgot a `Shutdown` command to shut all the layers down! I'm going to add this to the docs now. The idea is it propagates from the top down (app to filter to transport), causing everything to exit ASAP. We might want the server app layer in particular to have another `GentleShutdown` which notifies players and gives them time to finish their game while not allowing any new connections.

I also need to start adding some of the interface between filter and app layers.

Manghi:
The test case `basic_server_filter_flow()` does not expire on my system. It hangs indefinitely. Does it work for you?

```bash
....
conwayste/netwaystev2 $ cargo test 
test filter::tests::sortedbuffer_tests::seqminheap_insert_sequential_with_gaps ... ok
test filter::tests::time_advance_tests::time_advancing_works ... ok
test filter::tests::time_advance_tests::basic_server_filter_flow has been running for over 60 seconds
^C
conwayste/netwaystev2 $ 
```

AaronM04:
> The test case `basic_server_filter_flow()` does not expire on my system. It hangs indefinitely. Does it work for you?
> 
> ```shell
> ....
> conwayste/netwaystev2 $ cargo test 
> test filter::tests::sortedbuffer_tests::seqminheap_insert_sequential_with_gaps ... ok
> test filter::tests::time_advance_tests::time_advancing_works ... ok
> test filter::tests::time_advance_tests::basic_server_filter_flow has been running for over 60 seconds
> ^C
> conwayste/netwaystev2 $ 
> ```

I ran it several times and sometimes it fails while other times it hangs forever. I was running the wrong test (`cargo test -p netwaystev2 time_advance_test`) and got excited that it was passing and not hanging.  I'll take a look tonight.

netwaystev2/src/filter/filter.rs
```diff
@@ -0,0 +1,756 @@
...
+
+        let filter_cmd_rx = self.filter_cmd_rx.take().unwrap();
+        let filter_rsp_tx = self.filter_rsp_tx.clone();
+        let filter_notice_tx = self.filter_notice_tx.clone();
+        tokio::pin!(filter_cmd_rx);
+        tokio::pin!(filter_rsp_tx);
+        tokio::pin!(filter_notice_tx);
+
+        let mut ping_interval_stream = tokio::time::interval(Duration::new(2, 0));
+        ping_interval_stream.tick().await;
```
AaronM04, line 159: @Manghi adding this line causes the `basic_server_filter_flow` test to fail.

AaronM04 (commented):
plz fix

netwaystev2/src/filter/filter.rs
Manghi: Fixed.

AaronM04: Nice catch with the `tokio::task::yield_now`!

netwaystev2/src/filter/filter.rs
```diff
@@ -422,6 +425,17 @@ impl Filter {
...
                     .expect("sequence number cannot be None by this point"); // expect OK because of above check
                 while let Some(response_code) = server.response_codes.take_if_matching(expected_seq_num.0) {
+                    // When joining or leaving a room, the game_updates are reset
+                    match response_code {
+                        ResponseCode::JoinedRoom { .. } => {
+                            server.game_update_seq = None;
+                        }
+                        ResponseCode::LeaveRoom => {
+                            server.game_update_seq = None;
+                        }
```
Manghi: You should be able to condense it via an  `|` operator.

netwaystev2/src/filter/filter.rs
```diff
@@ -451,17 +465,87 @@ impl Filter {
...
+                                    diff: gen_state_diff,
+                                    endpoint,
+                                })
+                                .await?;
+                        }
+                    }
+                }
+
+                let mut start_idx = None;
+                match (server.game_update_seq, game_update_seq) {
```
Manghi: Can you leave a comment explaining what the purpose of this block is? This seems like one of those things that makes sense now but will be easy to forget after several months.

netwaystev2/src/filter/filter.rs
```diff
@@ -451,17 +465,87 @@ impl Filter {
...
+                            ));
+                        }
+                    }
+                    for i in (_start_idx as usize)..game_updates.len() {
+                        if let Some(ref mut room) = server.room {
+                            if let Err(e) = room.process_game_update(&game_updates[i]) {
+                                error!("[FILTER] failed to process game update {:?}: {}", game_updates[i], e);
+                            }
+                        }
+                        server.game_update_seq.as_mut().map(|seq| *seq += 1); // Increment
```
Manghi: Comment is a bit useless. Perhaps explain why the increment is needed?

netwaystev2/src/filter/client_update.rs
```diff
@@ -1,29 +1,136 @@
...
 use crate::protocol::{GameUpdate, GenPartInfo, GenStateDiffPart, UniUpdate};
 use conway::{
     rle::Pattern,
-    universe::{GenStateDiff, Universe},
+    universe::{BigBang, GenStateDiff, PlayerBuilder, Region, Universe},
 };
 
 pub struct ClientRoom {
     player_name:       String, // Duplicate of player_name from OtherEndServer (parent) struct
+    player_id:         Option<usize>, // If player is not a lurker, must be Some(...) before the first GenStateDiff
```
Manghi, line 14: Would you be opposed to aliasing `Option<usize>` to `PlayerID`?

netwaystev2/src/filter/client_update.rs
```diff
@@ -1,29 +1,136 @@
...
+                    .history(options.history as usize)
+                    .fog_radius(options.fog_radius as usize);
+
+                for net_region in options.player_writable.iter() {
+                    let region: Region = net_region.clone().into();
+                    big_bang = big_bang.add_player(PlayerBuilder::new(region));
+                }
+                let uni = big_bang.birth()?;
+                let game = ClientGame {
+                    player_id:     self.player_id, // copy
```
Manghi, line 52: What's the comment for?

netwaystev2/src/filter/client_update.rs
```diff
@@ -1,29 +1,136 @@
...
+                    partial_gen:   None,
+                };
+                self.game = Some(game);
+            }
+            GameFinish { .. } => {
+                self.game = None;
+                self.player_id = None;
+                let mut new_other_players = HashMap::new();
+                for name in self.other_players.keys() {
+                    new_other_players.insert(name.clone(), None);
```
Manghi, line 65: Why does everyone become a lurker on `GameFinish`?

netwaystev2/src/filter/client_update.rs
```diff
@@ -1,29 +1,136 @@
...
+                }
+                self.other_players = new_other_players;
+            }
+
+            PlayerList { players } => {
+                for player in players {
+                    if player.name == self.player_name {
+                        // Hey, that's us!
+                        self.change_own_player_id(player.index);
```
Manghi, line 74: Why does the index change here on a `PlayerList` command? Is that if the server assigned the player a different Universe index?

AaronM04: Yes, that's the reason.

netwaystev2/src/filter/client_update.rs
AaronM04: It's copying the `player_id` from the room to the game.

netwaystev2/src/filter/client_update.rs
AaronM04: Just an arbitrary choice. We can revisit when exploring the actual game play. :)

netwaystev2/src/transport/endpoint.rs
```diff
@@ -0,0 +1,272 @@
...
+
+    /// Returns a list of packets that can be retried across all endpoints.
+    /// Side effect: updates last_transmit and retry_count on any packets that can be retried.
+    pub fn retriable_packets(&mut self) -> Vec<(&P, Endpoint)> {
+        let mut retry_qualified = vec![];
+
+        for (endpoint, container) in &mut self.transmit {
+            for PacketContainer { packet, info, tid } in container {
+                // Add the packet to the list of retriable packets if enough time has passed since the last transmission
+                if info.transmit_interval != Duration::ZERO
```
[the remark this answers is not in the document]
AaronM04: Well then it would keep sending it as fast as possible. :smile:

nwv2-python-wrapper/src/protocol/update.rs
```diff
@@ -0,0 +1,54 @@
+use pyo3::exceptions::*;
```
migdall, line 1: Is that a Rust import?

AaronM04: It is! It's importing from this: https://docs.rs/pyo3/latest/pyo3/exceptions/index.html

These are Rust wrappers for some standard Python exception classes like `Exception`, `ValueError`, etc.

Also hi 👋

PR #282, opened 2022-11-16
Fix build issues with SemVer package update on master

Manghi:
- Alias the internal error as the previous `SemVerError`
- `if let` ladder to pull out versions for comparison


Manghi:
Compiles on my machine with rust `v1.67.0-nightly`

AaronM04 (approved):

PR #318, opened 2023-06-01
Remove old dissect-netwayste (Wireshark plugin)

AaronM04:
We now have a libpcap-based replacement in the conwayste-tools repository.


Manghi (approved):
Huzzah! That code I wrote was awful.

PR #319, opened 2023-06-01
Remove ggez, rodio, cpal, and gilrs submodules

AaronM04:
These submodules were being used for running this on OpenBSD, which (for now) we will not try to support.

Note that this is a _slightly_ different version of ggez (0.6.1, previously 0.6.0-rc1).

For the changes in this commit to take effect, run `git submodule deinit --all`. To undo (for example, if switching to another branch that doesn't have this commit), run `git submodule update --init --recursive`.


Manghi (approved):
Another huzzah!

PR #323, opened 2023-06-11
GUI: fix bug where buttons incorrectly selected on screen change

AaronM04:
You can see this in the master branch by starting at the main menu and activating either the Start Single Player Game or Options buttons by either keyboard or mouse, and then pressing the Esc key to go back. The keyboard and mouse activation issues are actually two separate bugs, both of which I fixed in this PR. :grinning:


Manghi (approved):
LGTM

